PR #15, opened 2014-12-12
Functions for easy coloring

jcheng5:
Adds a family of functions for mapping values to colors.

|  | Continuous values | Categorical values |
| --- | --- | --- |
| Continuous colors | `colorNumeric` | N/A |
| Discrete colors | `colorBin` and `colorQuantile` | `colorFactor` |

This makes it possible to describe most color mappings in a single function call.

Hat tip to @hadley, who has forgotten more than I will ever know about scaling.

I would love some eyes on this now but there are some open questions.
- [x] What happens when NA is fed in? `colorNumeric(...)(NA)` Should that return NA, or (optionally) you provide `colorNumeric` with an NA color?
- [x] What happens for `colorNumeric`, `colorBin`, and `colorQuantile` if a value is fed in that is too low/too high for the domain? Snap to the nearest valid value, with a warning? Return NA? Provide an option?
- [x] What happens for `colorFactor` when a value is fed in that isn't one of the expected levels? Return NA? Return an NA color?
- [ ] `cut` takes `include.lowest` and `right` options, I was thinking about exposing them on `colorBin` but reversing the defaults.

### DIFF
--- a/inst/htmlwidgets/leaflet.js
+++ b/inst/htmlwidgets/leaflet.js
@@ -235,13 +235,12 @@ var dataframe = (function() {
       (function() {
         var popup = L.popup(df.get(i))
                      .setLatLng([df.get(i, 'lat'), df.get(i, 'lng')])
-                     .setContent(df.get(i, 'content'))
-                     .openOn(this);
+                     .setContent(df.get(i, 'content'));
         var thisId = df.get(i, 'layerId');
         this.popups.add(popup, thisId);
-        marker.on('click', mouseHandler(this.id, thisId, 'marker_click'), this);
-        marker.on('mouseover', mouseHandler(this.id, thisId, 'marker_mouseover'), this);
-        marker.on('mouseout', mouseHandler(this.id, thisId, 'marker_mouseout'), this);
+        popup.on('click', mouseHandler(this.id, thisId, 'popup_click'), this);
+        popup.on('mouseover', mouseHandler(this.id, thisId, 'popup_mouseover'), this);
+        popup.on('mouseout', mouseHandler(this.id, thisId, 'popup_mouseout'), this);
       }).call(this);
     }
   };


barryrowlingson:
Still don't see an example of this integrated with the mapping functions. Currently I'm doing:

```
# define a palette function:
cs = colorQuantile(rev(brewer.pal(5,"Spectral")),nc.sids$SID79,n=5)
# map
leaflet(data=nc.sids) %>% addTiles() %>% addPolygons(fillColor=~cs(SID79),fillOpacity=.7)
```
(data from example(nc.sids) from the spdep package)

Is that the best pattern?

jcheng5:
That looks right to me--is it working?

FYI, if you are always going to use the same exact data to set up colorQuantile as you're going to pass into the resulting color function, you could also create colorQuantile with `domain = NULL`. When you do that, the domain will be whatever data is encountered when coloring is applied.

barryrowlingson:
Yup, it works fine. I had just worked out how to persuade ggplot2 scale_* functions to give up their secretes using some unexported functions but this is just as good.

johnpauls:
These functions are great. It would be nice if there was a way to create a legend using the same colors and display it on the map - somewhat alluded to in #32 .

jcheng5:
Yes sorry, the lack of an easy legend is a gaping hole that we want to plug soon.

mbacou:
This is useful, but why not just wrap all of Bivand's classInt::classIntervals() classification methods instead ("fixed", "sd", "equal", "pretty", "quantile", "kmeans", "hclust","bclust", "fisher", and "jenks")? I work with biophysical data that often looks best with the "kmeans" method for example, and "jenks" tend to be geographers' favorites.

jcheng5:
@mbacou That's a nice idea--maybe as a separate `colorInterval` function?

I think right now if you pass `colorBin(bins = classIntervals(...)$brks)` that should work, can you correct me if I'm wrong?

mbacou:
@jcheng5 yes you're correct. My bad, I wrongly assumed `colorBin()` was for equal interval bins, not just any interval. In that case a `colorInterval()` wrapper would be nice (save some key strokes), but not essential.

hadley:
In my opinion, there's very little evidence that more complex options have significant improvements over equal and quantile. I don't think they should be baked in to leaflet, given how easy they are to get from classIntervals.

Issue #29, opened 2015-01-09
Click event handlers are added but not cleared


wch:
A `click` event handler is added here each time the map receives an update, but it isn't cleared. This results in the data being sent to the server more and more times.

https://github.com/rstudio/leaflet/blob/8353463bc232795d97d6dcc116f2bf82eccdf841/inst/htmlwidgets/leaflet.js#L529-L535

The same may be true for the `moveend` event a few lines below.

Edit: I think that `shiny.onInputChange` may collapse multiple calls into a single data item sent to the server. But `shiny.onInputChange` is still called an increasing number of times as more clicks occur.

jcheng5:
Whoops.

The nonce should prevent the multiple calls from being collapsed, it'd be weird if you're not seeing multiple messages being sent to the server.

Do you mind taking care of these while you're in there adding "if first time rendering then..." logic?

wch:
The `onInputChange` still collapses them into a single thing to send to the server. I think it's because they all happen before the next `setTimeout(0)`.

https://github.com/rstudio/shiny/blob/f066a4bdf495f1ddb6a111ee0fe33a967c7c682c/inst/www/shared/shiny.js#L356-L387

PR #30, opened 2015-01-13
Add zoomToLimits function

wch:
This adds a new function called `zoomToLimits`.

By default, when objects are added to a map, when the map renders, it zooms to fit the objects (this behavior is unchanged from before):

```R
# Zooms to objects (default)
m <- leaflet() %>% 
  addTiles() %>%
  addPopups(174.7690922, -36.8523071, 'R was born here!')
m

# Same effect with `zoomToLimits(TRUE)`:
m %>%  zoomToLimits(TRUE)

# Don't zoom to the objects:
m %>%  zoomToLimits(FALSE)
```

This is useful in Shiny applications, where the data can be updated and you don't want the map to re-zoom. The following app is an example of it in use. In the following app:

* The data changes every 5 seconds.
* On the first rendering, the map zooms to the data.
* On subsequent data updates, the map doesn't re-zoom. If you zoom manually, it won't re-zoom when the data updates.
* There's a button to re-zoom to the data objects.

Live demo: https://winston.shinyapps.io/rezoom-leaflet/

```R
library(leaflet)

ui <- fluidPage(
  sidebarLayout(
    sidebarPanel(
      actionButton("zoomButton", "Zoom to objects")
    ),
    mainPanel(
      leafletOutput("map", height = 500)
    )
  )
)


  # Keep track of whether map has rendered yet
  mapHasRendered <- FALSE
  # Track last value of zoomButton actionbutton
  lastZoomButtonValue <- NULL

  output$map <- renderLeaflet({
    # New data every 5 seconds
    invalidateLater(5000, session)

    m <- leaflet() %>%
      addTiles() %>%
      addCircleMarkers(rnorm(50, -93.65, .01), rnorm(50, 42.0285, .01))


    rezoom <- FALSE
    # If zoom button was clicked this time, and store the value, and zoom to objects
    if (!identical(lastZoomButtonValue, input$zoomButton)) {
      lastZoomButtonValue <<- input$zoomButton
      rezoom <- TRUE
    }
    # Zoom to objects if map hasn't yet rendered for the first time
    if (!mapHasRendered) {
      mapHasRendered <<- TRUE
      rezoom <- TRUE
    }

    m %>% zoomToLimits(rezoom)
  })
}

shinyApp(ui, server)
```

The code for the app is admittedly kind of clunky. I think the logic for checking whether the `actionButton` was clicked is unavoidable. A different approach to this problem could potentially eliminate the need for the  `mapHasRendered` check, but at the expense of more logic and state on the Javascript side.


jcheng5:
I'm OK with the code for the zoom button being a little clunky, but I don't like how clunky the mapHasRendered code is--that seems like a common case to me. What did you think of having three options for value that represent "always", "never", and "on initial rendering only"?

wch:
That sounds reasonable. To do it, we'll need a place to save state and options on the browser side, which could also affect how #29 is implemented. Would it be OK to add a new field to the `map` objects, like `map.leafletr`?

jcheng5:
Sure, but just to be clear, the only state you'd need to save at this time is whether the map has been rendered before, right?

PR #31, opened 2015-01-14
Handle click and bounds update events only once

### LINKED ISSUE #29: Click event handlers are added but not cleared

wch:
Fixes #29. Click and bounds-update event handlers will only be registered once per map. Previously, each time a data update was received from the server, an additional click and bounds-udpate event handler was added.

This also adds a set of functions for managing event handlers. Leaflet doesn't support event namespacing; the new functions provide that functionality.

### DIFF
--- a/R/leaflet.R
+++ b/R/leaflet.R
@@ -16,8 +16,6 @@
 #'   \code{Polygons}, \code{SpatialPolygons}, \code{SpatialPolygonsDataFrame},
 #'   \code{Line}, \code{Lines}, \code{SpatialLines}, and
 #'   \code{SpatialLinesDataFrame})
-#' @param id a character string as the identifier of the map (you do not need to
-#'   provide it unless you want to manipulate the map later in Shiny)
 #' @param width the width of the map
 #' @param height the height of the map
 #' @param padding the padding of the map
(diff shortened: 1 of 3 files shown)


jcheng5:
How about just moving the `.on` calls to `initialize()` rather than `renderValue()`?

wch:
That could work, provided that `mapId` is accessible from the `map` object, but it's not clear to me that that's the case.

jcheng5:
You can get the ID using `this.getId(el)`.

Oh... shoot. I just realized, leaflet.js is using the wrong ID. It shouldn't be grabbing data.mapId, it should be using `this.getId(el)` and the leaflet function shouldn't take an id argument at all. (cc @yihui)

yihui:
@jcheng5 I agree. Actually I removed the id argument from `leaflet()` in one of my local branches.

wch:
OK, I've redone this PR. I've removed `mapId`, and register the event handlers on initialization. It's much simpler this way.

yihui:
Looks good to me.

PR #33, opened 2015-01-20
Rezoom

wch:
This is a new implementation of #30. It adds a function called `mapOptions()`, which has an argument `zoomToLimits`. The value can be "always", "first", or "never".

he following app is an example of it in use. In the following app:

* The data changes every 5 seconds.
* On the first rendering, the map zooms to the data.
* On subsequent data updates, the map doesn't re-zoom. If you zoom manually, it won't re-zoom when the data updates.
* There's a button to re-zoom to the data objects.

Live demo: https://winston.shinyapps.io/rezoom-leaflet/

For this particular app, there's a little extra complication in getting the map to re-zoom when the button is clicked. However, this isn't something that most users will run into, and something we can address in the future with (possibly) an imperative interface to the map.

```R
library(leaflet)

ui <- fluidPage(
  sidebarLayout(
    sidebarPanel(
      actionButton("zoomButton", "Zoom to objects")
    ),
    mainPanel(
      leafletOutput("map1", height = 500)
    )
  )
)


  # Track last value of zoomButton actionbutton
  lastZoomButtonValue <- NULL

    # Re-run every 5 seconds
    invalidateLater(5000, session)

    m <- leaflet() %>%
      addTiles() %>%
      addCircleMarkers(rnorm(50, -93.65, .01), rnorm(50, 42.0285, .01))

    rezoom <- "first"
    # If zoom button was clicked this time, store the value, and zoom to objects
    if (!identical(lastZoomButtonValue, input$zoomButton)) {
      lastZoomButtonValue <<- input$zoomButton
      rezoom <- "always"
    }

    m %>% mapOptions(zoomToLimits = rezoom)
  })

}

shinyApp(ui, server)
```


R/leaflet.R
```diff
@@ -43,3 +43,33 @@ leaflet = function(data = NULL, width = NULL, height = NULL, padding = 0) {
...
+#'   behavior is \code{"always"}.
+#'
+#' @examples
+#' # Don't auto-zoom to the objects (can be useful in interactive applications)
+#' leaflet() %>%
+#'   addTiles() %>%
+#'   addPopups(174.7690922, -36.8523071, 'R was born here!') %>%
+#'   mapOptions(zoomToLimits = "first")
+#' @export
+mapOptions <- function(map, zoomToLimits = NULL) {
```
jcheng5: Isn't it the usual thing to have `zoomToLimits = c("always", "first", "never")` in the parameter declaration, and then `zoomToLimits <- match.arg(zoomToLimits)` down below?

jcheng5:
Other than my one nitpick, LGTM!

R/leaflet.R
wch: Good catch - I used a NULL default here because, in the previous incarnation, the options would be stored on the client side and I wanted to allow users to call this function without changing a particular option. But with the way it works now, that issue is moot.

Issue #35, opened 2015-01-26
Unable to use click on a geojson feature


mennodekker:
I added my map using the leafletOutput() and leaflet() functions. 
In my server.R I add a geoJson layer using:

map = map %>%   addGeoJSON(features)

The features show up fine, but I can not track a click on the feature. Some debugging learned that in the leaflet.js in the geoJSON method it tries to register a click on the self.id (mapid) but it is not present at that time so it registers an "undefined" mapid.

mbacou:
@mennodekker, for the mere mortal, could you provide a quick example on how to use the click event either on the server or client side (or both). I'm migrating from @jcheng5 leaflet to @rstudio leaflet package and having difficulties replicating the onclick and onmouseover events using this version. That'd be really helpful, thanks.

mennodekker:
I did the same thing. The problem I had was that for geojson layers the mapid is not set. The work around until it is fixed is to use undefined.

As snippet of code that works with current implentation:

For **server.R** (notice the normal map click uses myMap and geosjon undefined)
```
  Lmap = leaflet() %>% setView(5.266007882805511, 16.5234375, zoom = 1.5)

   observe({
      evt <- input$myMap_click
      if (is.null(evt))
        return()
    
      isolate({
        # An empty part of the map was clicked.
        # Null out the selected feature.
        values$selectedFeature <- NULL
      })
    })
  
    observe({
      evt <- input$undefined_geojson_click
    
      if (is.null(evt))
        return()
    
      isolate({
        # A GeoJSON feature was clicked. Save its properties
        # to selectedFeature.
        values$selectedFeature <- evt$properties
      })
    })

  Lmap = Lmap %>% addGeoJSON(geo)
  output$myMap <- renderLeaflet(Lmap)
```
For **ui.R**
```
leafletOutput('myMap', width="100%", 380)
```

mbacou:
Perfect, thanks so much!

PR #36, opened 2015-01-26
Fix for issue #35

mennodekker:
I know this should be fixed somewhere else, but this is as far as my understanding of what's happening goes. This fix works for me.


yihui:
Thanks! This should be fixed now by #38.

Issue #37, opened 2015-01-30
Available use of leaflet event


bthieurmel:
Hi,
Is there a way to use leaflet event via the package or do you think about integrate that in the futur ?
It can be very usefull !

jcheng5:
Can you tell us a little more about what you're trying to accomplish? It's possible we're already exposing the events you want. Thanks!

bthieurmel:
Yes of course ! I want for example can use the zoomend methods for enabled or not some data per zoom level
```
map.on('zoomend', onZoomend);
function onZoomend(){
...
}
```

jcheng5:
You can do this if you're using it with Shiny. If your leafletOutput has an output ID of "foo" then you can refer to `input$foo_zoom`. There's also `input$foo_bounds` which is a named list with `north`/`south`/`east`/`west` elements.

bthieurmel:
Ok, thanks. And so, if I understand, there is no way actually to do this using in the viewer or saving to html file. Do you plan to explore this features or not ?

pssguy:
@jcheng5 

How would I do this in rmarkdown shiny?  I use the  `zoomToLimits="first" `option but would like the initial map
to be zoomed in 1 or 2 steps (it currently shows around 3 sets of the world)

    output$map <- renderLeaflet({

    leaflet( blah blah blah) %>% 
    mapOptions(zoomToLimits="first")

    })

    leafletOutput("map")

PR #40, opened 2015-02-04
Feature/events

bthieurmel:
Linked with  Available use of leaflet event #37 issue, can we think about something like that to enable user to set all event he want ?

http://leafletjs.com/reference.html#map-click

Then, we can do with a basic example : 

```
m = leaflet() %>% addTiles() %>% setView(-93.65, 42.0285, zoom = 17) %>%
 addEventListener(click ="function(e) {alert('click');}",
 zoomend ="function(e) {alert('zoomend');}")
```

addEventListener is actually write very quickly just to test, we can add more verificiations...

### DIFF
--- a/R/events.R
+++ b/R/events.R
@@ -1,0 +1,25 @@
+#' Methods to add event listener map widget
+#'
+#' A series of methods to manipulate the map.
+#' @param map a map widget object created from \code{\link{leaflet}()}
+#' @param ... any arguments with names equal to event name, and value to javascript function. See example.
+#' @references \url{http://leafletjs.com/reference.html#map-click}
+#' @return The modified map widget.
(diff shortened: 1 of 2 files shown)


bhaskarvk:

jcheng5:
I don't think we should add this so late in the v1.1 cycle. You can achieve this (admittedly with a little more typing) using onRender and I'd want to think carefully about such a feature. Just off the top of my head:

1. `addEventListener` feels like too general a function name.
2. `onRender` gives you access to the map object, element, and arbitrary data. I think those might be useful here too.
3. How to unsubscribe?

bhaskarvk:
OnRender is good enough, so closing this.

PR #48, opened 2015-03-25
specify an icon using an R list

yihui:
This is an implementation of specifying an icon (or a list of icons) using a list in R, and create it on the JS side using `L.icon()`.

Some examples:

```r
m = leaflet() %>% addTiles()
m %>% addMarkers(-87.9034077, 41.9728589, icon = list(
    iconUrl = 'http://i.imgur.com/zq3z3e3.png',
    iconSize = c(80, 40)
))

Rlogo = file.path(R.home("doc"), "html", "logo.jpg")

m = leaflet() %>% addTiles()
m %>%
  addMarkers(174.7690922, -36.8523071, icon = list(
    iconUrl = Rlogo,
    iconSize = c(100, 76)
  )) %>%
  addPopups(174.7690922, -36.8523071, 'R was born here!')

m %>% addMarkers(rnorm(30, 175), rnorm(30, -37), icon = list(
  iconUrl = Rlogo,
  iconSize = c(25, 19)
))
```

screenshot from 2015-03-25 00 21 31

```r
m %>% addMarkers(
  c(-71.0382679, -122.1217866), c(42.3489054, 47.6763144), icon = list(
    iconUrl = 'http://www.rstudio.com/wp-content/uploads/2014/03/blue-125.png'
  ), popup = c('RStudio @ Boston', 'RStudio @ Seattle')
)
```

screenshot from 2015-04-16 00 19 47

```r
iconData = data.frame(
  lat = c(rnorm(10, 0), rnorm(10, 1), rnorm(10, 2)),
  lng = c(rnorm(10, 0), rnorm(10, 3), rnorm(10, 6)),
  group = rep(sort(c('green', 'red', 'orange')), each = 10),
  stringsAsFactors = FALSE
)

leaflet() %>% addMarkers(
  data = iconData,
  icon = ~ iconList(
    iconUrl = sprintf('http://leafletjs.com/docs/images/leaf-%s.png', group),
    shadowUrl = 'http://leafletjs.com/docs/images/leaf-shadow.png',
    iconSize = list(c(38, 95)), shadowSize = list(c(50, 64)),
    iconAnchor = list(c(22, 94)), shadowAnchor = list(c(4, 62)),
    popupAnchor = list(c(-3, -76))
  )
)
```

screenshot from 2015-04-16 00 20 12

```r
# use point symbols from base R graphics as icons
pchIcons = function(pch = 0:14, width = 30, height = 30, ...) {
  n = length(pch)
  files = character(n)
  # create a sequence of png images
  for (i in seq_len(n)) {
    f = tempfile()
    png(f, width = width, height = height, bg = 'transparent')
    par(mar = c(0, 0, 0, 0))
    plot.new()
    points(.5, .5, pch = pch[i], cex = min(width, height) / 8, ...)
    dev.off()
    files[i] = f
  }
  files
}

iconData = matrix(rnorm(500), ncol = 2)
res = kmeans(iconData, 10)
iconData = cbind(iconData, res$cluster)
colnames(iconData) = c('lat', 'lng', 'group')
iconData = as.data.frame(iconData)

# 10 random point shapes for the 10 clusters in iconData
shapes = sample(0:14, 10)
iconFiles = pchIcons(shapes, 40, 40, col = 'steelblue', lwd = 2)

# note the data has 250 rows, and the icon list is of length 10; they are
# connected by the `group` variable: the i-th row of iconData uses the
# group[i]-th icon in the icon list
leaflet() %>% addMarkers(
  data = iconData,
  icon = iconList(
    iconUrl = iconFiles,
    popupAnchor = list(c(20, 0))
  ),
  iconGroup = ~ group,
  popup = ~ sprintf(
    'lat = %.4f, long = %.4f, group = %s, pch = %s', lat, lng, group, shapes[group]
  )
)
```

screenshot from 2015-04-16 00 20 50

Cc @jcheng5


jread-usgs:
@yihui would this support vector icons? (i.e., svgs?)

yihui:
@jread-usgs Yes, any image format is okay.

yihui:
@jcheng5 This PR is basically ready, except that I have not decided which package to use for base64 encoding. We have such functionality in several packages, such as base64, base64enc, markdown, httpuv, caTools, and so on.

jcheng5:
I'm happy with everything except the `iconGroup` syntax. We can go with it for now if we need to but I'd like to have input from @hadley before we go to CRAN.

R/layers.R
```diff
@@ -177,24 +177,125 @@ popupOptions = function(
   )
 }
 
-#' @param icon the icon for markers; if you want to create a new icon using
-#'   JavaScript, please remember to use \code{\link[htmlwidgets]{JS}()} on the
-#'   JavaScript string; see \url{http://leafletjs.com/reference.html#icon}
+#' @param icon the icon(s) for markers; you can either use literal JavaScript
+#'   code (e.g. \code{\link[htmlwidgets]{JS}('L.icon({iconUrl: '?', size: [?,
+#'   ?]})')} or an R list to create an icon (e.g. \code{list(iconUrl = '?', size
+#'   = c(x, y))}), or use \code{\link{iconList}()} to create a list of icons;
```
jcheng5: Let's clean up the way size is specified for multiple icons

R/layers.R
```diff
@@ -177,24 +177,125 @@ popupOptions = function(
...
+#'   the HTML page so the icon images will still be available even when you
+#'   publish the map elsewhere (use \code{iconList(..., embed = FALSE)} if you
+#'   do not want the images to be encoded and embedded)
+#' @param iconGroup when \code{icon} contains a list of different icons, this
+#'   argument is a numeric vector of the same length as \code{lat} and
+#'   \code{lng}, where \code{iconGroup[i]} is the index of the icon in the
+#'   \code{icon} list for \code{lat[i]} and \code{lng[i]}; the main purpose of
+#'   this argument is to reduce the size of the icon data when the icon images
+#'   are base64 encoded: it is not efficient to encode the same image multiple
+#'   times when the image is used multiple times on the map, and we can encode
```
jcheng5: It would be great to also support deduplication of image data on the actual `icon` parameter as well.

inst/examples/iconList.R
```diff
@@ -0,0 +1,66 @@
...
+)
+
+leaflet() %>% addMarkers(
+  data = iconData,
+  icon = ~ iconList(
+    iconUrl = sprintf('http://leafletjs.com/docs/images/leaf-%s.png', group),
+    shadowUrl = 'http://leafletjs.com/docs/images/leaf-shadow.png',
+    iconSize = list(c(38, 95)), shadowSize = list(c(50, 64)),
+    iconAnchor = list(c(22, 94)), shadowAnchor = list(c(4, 62)),
+    popupAnchor = list(c(-3, -76))
```
jcheng5: Let's try to stay away from lists of x-y or lat-lng vectors. Either `iconAnchorX/iconAnchorY` or add a helper function that takes x and y or lat and lng.

hadley:
@jcheng5 anything in particular you want comment on?

yihui:
@hadley We feel the `iconGroup` syntax is kind of awkward, and wonder if you have a better idea about how to map a list of unique icons (say, of length m) to a vector of locations (of length n, and m < n). Currently I'm using an argument `iconGroup` to do that. For example, the list of icons may be `list(icon1, icon2, icon3)`, and `iconGroup` may be `c(1, 3, 3, 2, 1, 1, 1)` to index the icon list to match up with data/lat/lng. Normally the length of the icon list should be the same as `nrow(data)`, but we are doing base64 encoding here and do not want to duplicate the base64 data.

R/layers.R
hadley: I think it's a bit confusing to allow either js or R code. Why not have a small helper function like `leaftlet_icon <- function(url, width, height) {}`.  Then instead of `iconList()` you'd just use a (named) list of icons

R/layers.R
hadley: I think about this in a rather different way - I think you are creating a scale that maps between a categorical variable and a list of shapes. I think it would be better to call this argument `icon` and allow it to take three possible values:

* an icon object (in which case the same icon is used for everything)
* an integer vector (which indexes into the icon list)
* a character vector (which indexes into the icon list by name)

R/layers.R
hadley: Oh, looking at `iconList()` I think you'd still need it, but it would return a named list.

R/layers.R
jcheng5: Talked to @hadley over IM and we landed on a single argument, no `iconGroup`, and perform deduping. We can discuss over vchat if you want when you circle back to this.

yihui: @jcheng5 I think I can do deduping. 

@hadley I didn't get it for the latter two cases: where do you specify the icon list? You only mentioned the `icon` argument as an index vector, and I don't see the icon list to be indexed.

R/layers.R
```diff
@@ -237,12 +241,120 @@ addMarkers = function(
...
+#' possible attributes of icons.
+#'
+#' Note some icon attributes are of length 2, such as \code{iconSize} and
+#' \code{iconAnchor}, and \code{iconList()} provides two separate arguments for
+#' these attributes, e.g. \code{iconSizeX = 20} and \code{iconSizeY = 40} means
+#' \code{iconSize = c(20, 40)} internally. Shorter argument values will be
+#' re-cycled. \code{NULL} values for these arguments will be ignored.
+#' @param iconUrl the URL to the icon image
+#' @param iconRetinaUrl the URL to a retina sized version of the icon image
+#' @param iconSizeX,iconSizeY size of the icon image in pixels (\code{iconSizeX}
```
jcheng5: Shouldn't these just be `iconWidth`/`iconHeight`?

R/layers.R
```diff
@@ -237,12 +241,120 @@ addMarkers = function(
...
+#'   is width and \code{iconSizeY} is height)
+#' @param iconAnchorX,iconAnchorY the coordinates of the "tip" of the icon
+#'   (relative to its top left corner, i.e. the top left corner means
+#'   \code{iconAnchorX = 0} and \code{iconAnchorY = 0)}, and the icon will be
+#'   aligned so that this point is at the marker's geographical location
+#' @param shadowUrl the URL to the icon shadow image
+#' @param shadowRetinaUrl the URL to the retina sized version of the icon shadow
+#'   image
+#' @param shadowSizeX,shadownSizeY size of the shadow image in pixels
```
jcheng5: Similarly, `shadowWidth`/`shadowHeight`? BTW, typo: `shadown`

R/layers.R
```diff
@@ -237,12 +241,120 @@ addMarkers = function(
...
+#'
+#' An icon can be represented as a list of the form \code{list(iconUrl,
+#' iconSize, ...)}. This function is vectorized over its arguments to create a
+#' list of icons. See \url{http://leafletjs.com/reference.html#icon} for the
+#' possible attributes of icons.
+#'
+#' Note some icon attributes are of length 2, such as \code{iconSize} and
+#' \code{iconAnchor}, and \code{iconList()} provides two separate arguments for
+#' these attributes, e.g. \code{iconSizeX = 20} and \code{iconSizeY = 40} means
+#' \code{iconSize = c(20, 40)} internally. Shorter argument values will be
```
jcheng5: I don't know who this sentence ("Note some icon attributes are of...") is written for. Can we just drop it?

R/layers.R
```diff
@@ -237,12 +241,120 @@ addMarkers = function(
...
+  attrs = as.data.frame(do.call(cbind, attrs))
+  group = apply(as.matrix(attrs), 1, paste, collapse = '\r') # one row to one value
+  group = as.integer(factor(group, unique(group)))  # map rows to groups
+  attrs = attrs[!duplicated(group), , drop = FALSE] # dedup
+  structure(
+    lapply(seq_len(nrow(attrs)), function(i) {
+      L.icon(iconData(attrs[i, , drop = FALSE]), embed = embed)
+    }),
+    iconGroup = group
+  )
```
jcheng5: I'm quite confused by this structure, it seems both more complex and less efficient (bandwidth-wise) than the encoding we use for all the other functions--I think we use "wide" everywhere else and do recycling in the client, while you're using "long" here and doing... some recycling now, some recycling later?

R/layers.R
yihui: That makes sense.

R/layers.R
yihui: It was written for those who read the API reference http://leafletjs.com/reference.html#icon and I was concerned that they might be confused by the inconsistency of argument names, e.g. there is no `iconWidth`/`iconHeight` in the leafletjs documentation. Well, I guess it will be my problem if users really have to read the leafletjs documentation to understand these arguments, so I'll drop this sentence.

R/layers.R
yihui: Thanks!

R/layers.R
yihui: Yeah, this does look complicated. I thought quite a while about it. The two major issues that make it tricky to recycle icon arguments are:

1. The `icon` argument of `L.marker()` is at the same level as other marker options such as `clickable` and `draggable`, and `iconUrl`/`iconSize`/... are arguments of `L.icon()`. I'm not sure if you want to put all icon arguments into the `options` argument of `addMarkers()`, which will raise them to the same level as `clickable` after `dataframe.cbind(options)` in the client.
2. Recycling of icon arguments are different with other marker options; it is not simply `i % nrow`, but we will have to store the original indices of `iconUrl` somewhere (e.g. in the current PR, I'm using `iconGroup`), and index the deduped `iconUrl`.
  - For efficiency, we need to dedup all four URLs `'iconUrl', 'iconRetinaUrl', 'shadowUrl', 'shadowRetinaUrl'`, and store the original indices somewhere;
  - For the current PR, I did not dedup them separately, but was only looking at the unique combinations of the four URLs (each unique combination was base64 encoded); this is less efficient than base64 encoding them separately and store four vectors of indices;

I feel a little lost here. Perhaps we can chat tomorrow.

R/layers.R
```diff
@@ -237,12 +241,124 @@ addMarkers = function(
...
+  if (!is.list(options)) return(options)
+  if (is.null(names(options))) {
+    if (length(options) == 0) return()
+    # in theory we need to check all sub-lists and all must be named
+    if (is.null(names(options[[1]])))
+      stop('The data for an individual icon must be a named list')
+    # it is already a list of icons, so no further base64 encoding
+    return(options)
+  }
+  image_uri = getFromNamespace('.b64EncodeFile', 'markdown')
```
jcheng5: Ummm... I didn't realize you meant a *private* function from markdown... :)

Come to think of it, I might be bringing in Rcpp and a slew of native code for raster support, maybe we just copy in a base64 encoding routine while we're at it. Or is there a reason you didn't go with base64enc?

yihui: Yes, base64enc requires users to provide the mime info. I did not want to write a similar wrapper function like `markdown:::.b64EncodeFile` to make use of `mime::guess_type()`. Since the private function is in our own hands, I'm not too worried about it. 

I think it is a good idea if you can just bring in Rcpp, provided that we are going to use C++ anyway in the future. Thanks!

jcheng5:
This ended up being quite different--I did away with `L.icon` entirely. Mainly icons is just treated like another, independent bucket of options, with the addition of the packing/unpacking mechanism which is used for deduping.

yihui:
Okay, I just reviewed your code, and it does look much clearer than mine. Thanks a lot!

yihui:
@jcheng5 If there is not anything else you want to add, I can rebase this branch against master and merge.

jcheng5:
Please do go ahead and rebase against master. And actually if you come up with a better name for iconList or really don't want to change it you can go ahead and merge. We can figure out the last scenario independently I think.

yihui:
Okay, let's use `icons()` instead of `iconList()`, and talk about the third scenario tomorrow.

PR #53, opened 2015-04-01
color legends

yihui:
When a color palette function is created, some metadata is stored as the attributes of the function, which will be used to derive the legend items later in `addLegend()`.

Legends for factor variables are straightforward. I have not decided how to deal with numeric variables in the legend. I don't think it is a good idea to use the binning methods in the color palettes, since it is more important for the bins to be human-readable in the legend. `pretty()` might make more sense than `cut()`. See http://rpubs.com/yihui/73236 for some quick examples.

Please ignore the old examples below.

---

A manual legend (users specify the colors and labels):

```r
leaflet() %>% addTiles() %>% addLegend(
  position = 'bottomright',
  colors = rgb(t(col2rgb(palette())) / 255), labels = palette()
)
```

screenshot from 2015-04-01 16 43 11

Derive legend info from the color palette function:

```r
library(maps)
mapState = map("state", fill=TRUE, plot=FALSE)
snames = unique(mapState$names)
snames[snames == 'new york:main'] = 'new york'
snames[snames == 'washington:main'] = 'washington'
snames[grep('^(new york|washington):', snames)] = NA
snames = gsub(':.+$', '', snames)
sarea = state.area[match(snames, tolower(state.name))]
pal = colorNumeric('YlOrRd', sarea)

leaflet(data = mapState) %>% addTiles() %>%
  addPolygons(fillColor = pal(sarea), stroke = FALSE, fillOpacity = 0.5) %>%
  addLegend(pal = pal, values = sarea, bins = 10)
```

screenshot from 2015-04-01 16 43 56

```r
sregion = state.region[match(snames, tolower(state.name))]
pal = colorFactor('Dark2', sregion)
leaflet(data = mapState) %>% addTiles() %>%
  addPolygons(fillColor = pal(sregion), stroke = FALSE, fillOpacity = 0.5) %>%
  addLegend(pal = pal, values = sregion)
```

screenshot from 2015-04-01 16 44 16

* [x] colorBin should make pretty breaks by default
* [ ] Some easy way to automatically create a legend for polygon etc. layers
* [x] Continuous scale gets unique legend
* [ ] color functions should return attribute with break info or whatever
- [ ] take a look at classInt


jcheng5:
cc @hadley @wch

jcheng5:
Looks like we're off to a good start! :)

When you say WIP do you mean you don't want this merged yet? Or that there will be other PRs after this?

Can we walk through the code together some time tomorrow?

yihui:
@jcheng5 I just mean this is not ready to be merged. Yep, let's talk tomorrow.

wch:
Looks like a nice start!

jcheng5:
* colorBin should make pretty breaks by default
* Some easy way to automatically create a legend for polygon etc. layers
* Continuous scale gets unique legend
* color functions should return attribute with break info or whatever

mbacou:
Happy to see this is moving forward as well. For legend labels, maybe provide a `print()` or similar method to the color palette function to print well-formatted human-readable labels, see for example classInt implementation of this:

```
## S3 method for class 'classIntervals'
print(x, digits = getOption("digits"), ..., 
 under="under", over="over", between="-", cutlabels=TRUE, unique=FALSE) 
```

jread-usgs:
Would be nice to treat units (when applicable) as first-class citizens of legends.

mbacou:
Agreed with @jread-usgs a legend print() method might also need a `title = "My data, date range, units"` argument (seems like a logical place for indicating units), else could make use of `print(x, digits = getOption("digits"), ..., under="under", over="over", between="-", unit="", cutlabels=TRUE, unique=FALSE)`

johnpauls:
+1 on being easily able to include a legend title and units

yihui:
@wch @hadley I changed the colorBin() function to use `pretty()` to determine the cut points, and the `colorNumeric()` palette will generate a continuous color gradient now. I'd love to know your comments. Here are some examples: http://rpubs.com/yihui/73236 Thanks!

hadley:
Looks good. Only one comment/question: is it possible to show NA in the legend only when there's an NA in the data?

yihui:
@hadley Yes, that is exactly what I do. In these examples, there are NA values, so NA is displayed there.

hadley:
@yihui oh I see - it's just a tiny region that has an NA so I didn't notice it.

wch:
Looking pretty good!

One tricky thing is that with a semi-transparent color overlay on the map, the colors don't quite match the legend. For example, in the 2nd and 3rd maps, it's not entirely clear what numeric value a particular state (like Texas) has.

Would it be possible to default the legend to the same color as the tile background, and then use the same opacity for the legend as is used in the choropleth? I'm not sure how best to handle different colors in other tilesets, though.

yihui:
I'm aware of the opacity issue (currently the legend uses fully opaque colors). Let me try to adjust the opacity and see how it looks. Thanks!

R/legend.R
```diff
@@ -0,0 +1,70 @@
...
+  type = 'unknown'; extra = NULL; na.color = NULL
+  formatNum = function(x) format(x, scientific = FALSE, big.mark = ',')
+
+  if (!missing(pal)) {
+
+    type = attr(pal, 'colorType', exact = TRUE)
+    args = attr(pal, 'colorArgs', exact = TRUE)
+    na.color = args$na.color
+
+    if (type == 'numeric' || type == 'bin') {
```
jcheng5: Split into two separate cases

R/legend.R
```diff
@@ -0,0 +1,70 @@
+#' @export
+addLegend = function(
+  map, position = c('topright', 'bottomright', 'bottomleft', 'topleft'),
+  pal, values, na.label = 'NA', bins = 7, colors, opacity = 0.5, labels,
+  title = deparse(substitute(values))
+) {
```
jcheng5: Validate inputs:

* If labels is provided, it must be a vector of the same length as colors
* `missing(pal) == missing(values)`, `missing(colors) == missing(labels)`, `missing(pal) != missing(colors)`
* bins will be ignored if colors and labels are provided

R/legend.R
```diff
@@ -0,0 +1,70 @@
+#' @export
+addLegend = function(
```
jcheng5: Add documentation, examples

R/legend.R
jcheng5: Add comments :)

R/legend.R
```diff
@@ -0,0 +1,70 @@
...
+
+    } else if (type == 'quantile') {
+
+      p = args$probs
+      n = length(p)
+      cuts = quantile(values, probs = p, na.rm = TRUE)
+      mids = quantile(values, probs = (p[-1] + p[-n]) / 2, na.rm = TRUE)
+      colors = pal(mids)
+      p = paste0(round(p * 100), '%')
+      cuts = sprintf('%s &ndash; %s', formatNum(cuts[-n]), formatNum(cuts[-1]))
```
jcheng5: Seems to be an extra space getting introduced after the ndash

R/legend.R
jcheng5: Add tests

inst/htmlwidgets/lib/leafletfix/leafletfix.css
```diff
@@ -5,3 +5,22 @@ img.leaflet-tile {
...
 }
+.info {
+    padding: 6px 8px;
+    font: 14px/16px Arial, Helvetica, sans-serif;
+    background: white;
+    background: rgba(255,255,255,0.8);
+    box-shadow: 0 0 15px rgba(0,0,0,0.2);
+    border-radius: 5px;
+}
+.legend {
```
jcheng5: Please add a (very) few pixels of space under title of legend

R/legend.R
jcheng5: Investigate under/over/between/units

jcheng5:
Does this include whatever you learned about `classInt`? Should we talk about that?

R/legend.R
```diff
@@ -0,0 +1,162 @@
...
+#'   \code{colors}
+#' @param labelFormat a function to format the labels derived from \code{pal}
+#'   and \code{values}
+#' @param title the legend title
+#' @example inst/examples/legend.R
+#' @export
+addLegend = function(
+  map, position = c('topright', 'bottomright', 'bottomleft', 'topleft'),
+  pal, values, na.label = 'NA', bins = 7, colors, opacity = 0.5, labels,
+  labelFormat, title = deparse(substitute(values))
```
yihui: labelFormat = list(units = ?, over, under, between, ...)

yihui:
Whew, I think this PR is finally complete now. I added the `labelFormat()` function.

R/legend.R
```diff
@@ -0,0 +1,204 @@
...
+
+  formatNum = function(x) {
+    format(
+      round(transform(x), digits), trim = TRUE, scientific = FALSE,
+      big.mark = big.mark
+    )
+  }
+
+  if (missing(type)) {
+    # look for 'pal' in the parent environment and automatically get the type
```
jcheng5: HA! Nice try!

Let me take a stab at something cleaner.

jcheng5:
I made the legend type much more explicit--let me know what you think.

If you're OK with that, I think we're ready to merge.

yihui:
That looks great. Thanks!

PR #61, opened 2015-04-24
Include coords for all mouse events

jcheng5:
Previously would only send lat/lng back when the event's target
had a getLatLng() method. Now if getLatLng() is not found, we
use the lat/lng of the mouse itself.


jcheng5:
@yihui any objection to me merging this?

yihui:
@jcheng5 No. It looks good to me.

PR #64, opened 2015-04-26
feature to add addControl, removeControl, clearControls

mplourde:
This is a pull request to add support for adding and removing leaflet controls with arbitrary html content. [See the demo here.](https://plourde.shinyapps.io/addControl-demo/addControl-demo.Rmd)

This branch adds the file **R/controls.R** with the following definition of `addControl`:

```{r}
addControl = function(map, 
    position=c('topleft', 'topright', 'bottomleft', 'bottomright'), 
    html, 
    controlId=NULL,
    classes=c('info', 'legend'),
    data=getMapData(map)) {

    position <- 
        if (missing(position)) position[1]
        else position

    classes <- paste(classes, collapse=' ')
    html <- as.character(html);

    appendMapData(map, data, 'addControl', position, html, controlId, classes)
}
```

with correspoding javascript method in **inst/htmlwidgets/leaflet.js**

```{javascript}
methods.addControl = function(position, html, controlId, classes) {
  var onAdd = function(map) {
    var div = L.DomUtil.create('div', classes);
    div.innerHTML = html;
    if (typeof controlId !== 'undefined' && controlId !== null) {
      div.setAttribute('id', controlId)
    }
    this._div = div;
    return this._div;
  };
  var Control = L.Control.extend({options: {position: position}, onAdd: onAdd})
  this.controls.add(new Control, controlId);
};
```

Additionally, I've added a new "store" for controls in **inst/htmlwidgets/leaflet.js**, `ControlStore`, that allows for their management. This store behaves similarly to the already existing `LayerStore`.

```{javascript}
  function ControlStore(map) {
    this._controlsNoId = [];
    this._controlsById = {};
    this._map = map
  }

  ControlStore.prototype.add = function(control, id) {
    if (typeof(id) !== 'undefined' && id !== null) {
      if (this._controlsById[id]) {
        this._map.removeControl(this._controlsById[id]);
      }
      this._controlsById[id] = control;
    } else {
      this._controlsNoId.push(control)
    }
    this._map.addControl(control);
  };

  ControlStore.prototype.remove = function(id) {
    if (this._controlsById[id]) {
      this._map.removeControl(this._controlsById[id]);
      delete this._controlsById[id];
    }
  };

  ControlStore.prototype.clear = function() {
    for (var i = 0; i < this._controlsNoId.length; i++) {
        this._map.removeControl(this._controlsNoId[i]);
    };
    this._controlsNoId = [];

    for (var key in this._controlsById) {
        this._map.removeControl(this._controlsById[key])
    }
    this._controlsById = {}
  }
```

In order to fully support adding and removing controls via `leaflet:::leafletController`, I've modified the HTML widget `initialize` function so that it saves the `map` variable generated by leaflet during the initialization process in an attribute of the binding named `_map`. It's necessary to have access to this variable to call leaflet functions, such as `removeControl`, on the map outside of the `initialize` and `renderValue` functions. 

To get this variable in the leaflet custom message handler used by `leafletController`, I changed the line `var map = document.getElementById(mapId);` to `var map = Shiny.shinyapp.$bindings[mapId].binding._map`. I'm not sure if this is the easiest way to do this. These changes to the initialize function and the custom message handler also enable usage of the existing `remove*` methods defined in **inst/htmlwidgets/leaflet.js**, which, as far as I can tell, are not fully yet supported.


jcheng5:
Thanks for the contribution! In order to accept a nontrivial code submission we'll have to ask for a Contributor License Agreement--we do this so we can relicense the code if that becomes necessary in the future, to avoid incidents like [this](http://bit.ly/1nxBjJ2).

If you email me at joe@rstudio.com I'll send you the CLA document. Thanks!

jcheng5:
Signed CLA received, thanks.

mplourde:
I've updated `methods.addControl` in **inst/htmlwidgets/leaflet.js** so that any inputs supplied as content to the control will register as Shiny input bindings.

mplourde:
This pull request is no longer consistent with the updated API in master. I'm going to fork a new branch. Now that `leafletController` is removed, is there a recommended way to update an existing map by ID, aside from simply making a direct call to the custom message handler 'leaflet-calls`?

jcheng5:
Yes, start with leafletProxy(). We don't have extensive docs written up yet but I can walk you through it over vchat if you like. Or you can leave it and I'll just do the refactor today myself, up to you. 





On Thu, Apr 30, 2015 at 8:28 AM -0700, "Matthew Plourde" <notifications@github.com> wrote:












This pull request is no longer consistent with the updated API in master. I'm going to fork a new branch. Now that leafletController is removed, is there a recommended way to update an existing map by ID, aside from simply making a direct call to the custom message handler 'leaflet-calls`?

jcheng5:
Basically your R function calls invokeMethod where it used to call appendMapData. And invokeMethod will be smart enough to do the right thing whether you've got a leaflet or leafletProxy instance.

mplourde:
Thanks, Joe. I've updated this branch to be consistent with the new API.

jcheng5:
I see two issues that might be worth addressing:

- [ ] Shiny.unbindAll really ought to be called right before the control is removed or replaced.
- [ ] Calling `as.character()` on tags objects renders the HTML, but throws away information about HTML/JS/CSS dependencies that may be attached. I think we can capture it like this:

```r
deps <- htmltools::resolveDependencies(htmltools::findDependencies(html))
map$dependencies <- c(map$dependencies, deps)
invokeMethod(map, ...)
```

Sorry for all the complexity--this really was the simplest we were able to make things while still supporting all the scenarios we needed!

mplourde:
Joe, with what should I call `Shiny.unbindAll` in this situation? `Shiny.unbindAll(this._controlsById[id]._div)` in `ControlStore.prototype.remove`, for example, doesn't seem to be sufficient. I'm not sure how to verify that the inputs in the control have been unbound, but this call doesn't seem to change `Shiny.inputBindings` or `Shiny.shinyapp.$inputValues` and the value of the button I tested this with is still available on the server. Thanks.

jcheng5:
You've probably got it right--unbinding an input doesn't affect either `Shiny.inputBindings` (that's the *types* of bindings that are registered, not the *instances* of bindings that are active) or `$inputValues` (removing an input doesn't automatically cause its value to become unavailable--their state is loosely coupled). Mostly it's about allowing inputs to deregister event handlers so they can be garbage collected.

But more importantly, `unbindAll` also affects outputs, and those really do need to be unbound or else an exception will be thrown if someone tries to bind an output with that ID (which is totally legitimate, even common).

Feel free to IM me via gtalk at joe@rstudio.com if you have further questions.

mplourde:
Joe, in addition to what is in the commit message, this commit also adds support for HTML dependencies in controls. Instead of adding them to the map's dependencies, I pass them to `methods.addControl`, which hands them off to `Shiny.renderHtml`:

```{javascript}
  methods.addControl = function(position, html, dependencies, controlId, classes) {
    var onAdd = function(map) {
      var div = L.DomUtil.create('div', classes);
      if (typeof controlId !== 'undefined' && controlId !== null) {
        div.setAttribute('id', controlId)
      }
      this._div = div;

      Shiny.unbindAll(this._div);
      Shiny.renderHtml(html, this._div, dependencies);
      Shiny.initializeInputs(this._div);
      Shiny.bindAll(this._div);

      return this._div;
    };
    var Control = L.Control.extend({options: {position: position}, onAdd: onAdd})
    this.controls.add(new Control, controlId, html);
  };
```

jcheng5:
This no longer seems to work for non-Shiny cases. Rather than make you wade through the broken bits, I've fixed it on my own branch.

I've also swapped the order of the `html` and `position` arguments, is that a problem?

jcheng5:
Can you review the commit I added on the `joe/feature/addControl` branch, and if you're OK with it, merge it into your branch? Then we can have @yihui review and merge the entire thing.

Thanks for your patience.

mplourde:
Joe, I merged your branch. Thanks.

R/controls.R
```diff
@@ -0,0 +1,38 @@
...
+addControl = function(map,
+    html,
+    position=c('topleft', 'topright', 'bottomleft', 'bottomright'),
+    controlId=NULL,
+    classes=c('info', 'legend'),
+    data=getMapData(map)) {
+
+    position <-
+        if (missing(position)) position[1]
+        else position
```
yihui: Is there a reason that you want to avoid `match.arg()` here? Normally I just do `position = match.arg(position)`.

R/controls.R
```diff
@@ -0,0 +1,38 @@
...
+#' @param classes extra CSS classes to append to the control
+#'
+#' @describeIn map-layers Add arbitrary HTML controls to the map
+#' @export
+addControl = function(map,
+    html,
+    position=c('topleft', 'topright', 'bottomleft', 'bottomright'),
+    controlId=NULL,
+    classes=c('info', 'legend'),
+    data=getMapData(map)) {
```
yihui: I'm not sure if the `data` argument is necessary here. Probably not. `addControl()` is not like other `add*` functions that need the `data` object to derive lat/lng and other argument values.

yihui: Just thought more about it. Let's keep this argument. There might be cases in which we construct `html` using `data`, although I don't have an example on my mind right now.

yihui:
I just reviewed this PR, and it looks pretty good to me. Thanks @mplourde!

mplourde:
@yihui, thanks. I've updated the branch to reflect your comments.

yihui:
Okay, this looks good to me to merge now. I'll see if @jcheng5 has more comments.

jcheng5:
Thanks Matt!

PR #65, opened 2015-04-28
Add leaflet-providers plugin

jcheng5:
Example:

```r
leaflet() %>% addProviderTiles("Stamen.Toner")
```


jcheng5:
This also sets a precedent for integrating Leaflet plugins.

wch:
FYI, to get this through CRAN these days, the DESCRIPTION file will need a `c(person(...), person(...))` format, and you'll need a LICENSE file that has the licenses of all included components. See [shiny](https://github.com/rstudio/shiny) and [shinydashboard](https://github.com/rstudio/shinydashboard) for examples. They're also very particular about all contributors being credited in the DESCRIPTION if possible.

jcheng5:
Thanks @wch, made those changes.

jcheng5:
@yihui I believe this is ready to merge now.

yihui:
Thanks! This looks great!

DESCRIPTION
```diff
@@ -9,6 +9,7 @@ Authors@R: c(
       person("Hadley", "Wickham", role = c("ctb")),
       person("jQuery contributors", role = c("ctb", "cph"), comment = "jQuery library"),
       person("Leaflet contributors", role = c("ctb", "cph"), comment = "Leaflet library"),
+      person("Leaflet Providers contributors", role = c("ctb", "cph"), comment = "Leaflet Providers plugin"),
```
wch: I think that they should be listed as `ctb` only. See the LICENSE:
  https://github.com/Leaflet/Leaflet/blob/master/LICENSE

jcheng5: I don't understand... What about the license makes them not cph?

wch: Oops, I was looking at Leaflet, not Leaflet providers. But the point still stands, for the previous line.

Also, I think that Vladimir Agafonkin should be `cph` and `ctb`, and CloudMade should be as `cph`.

wch: For Leaflet, the contributors aren't copyright holders, according to the LICENSE file. (Though I suppose that if they haven't signed a CLA, it's possible the LICENSE file doesn't correctly reflect the copyright holders.)

jcheng5: Oh, I see what you're saying. I'll make the change on master. Thanks

PR #70, opened 2015-05-01
Make icon set

yihui:
- I renamed makeIconSet() to iconList()
- I figured out the `evalFormula()` issue: `lapply(x)` discarded the class of x
- I guess the argument names `iconUrl`/`shadowUrl` are fine, and I did not change them
- I added a simple check in iconList() to make sure the icons are from `makeIcon()` by using a class `leaflet_icon`

cc @jcheng5


jcheng5:
Thanks for taking care of that!

Issue #71, opened 2015-05-04
addCircles could handle NA values for lng and lat arguments


hansthompson:
I'd like to have the default behavior of addCircles to ignore NA values.  For example:
```
library(leaflet)  #a super awesome package
leaflet_sample <- data.frame(lati = c(61.191, NA, 61.171), long = c(-149.895, NA, -149.875))

leaflet() %>% addTiles() %>% 
  setView(-149.885, 61.181, zoom = 11) %>% addCircles(lng = leaflet_sample$long, lat = leaflet_sample$lati,
                                                      color = "#ff0000")
#OR 
leaflet() %>% addTiles() %>% 
  setView(-149.885, 61.181, zoom = 11) %>% addCircles(lng = leaflet_sample$long[!is.na(leaflet_sample$long)], 
                                                      lat = leaflet_sample$lati[!is.na(leaflet_sample$lati)], 
                                                      color = "#ff0000")
```
Thanks for any opinions if this is a worth feature request or not.

yihui:
Interesting. I did not know NA's could cause problems in leaflet. Thanks for the minimal reproducible example! We will see what we can do.

rbdixon:
+1 -- I just stumbled on this same issue when using addMarkers(). This seems to already be well documented but I do have a minimal reproducible example of the addMarker() behavior, if needed.

bhaskarvk:
Closing as duplicate and tracking under #304

PR #72, opened 2015-05-04
Remove NA's from addCirlce's lng and lat arguments as the default.  

hansthompson:
Perhaps na.ignore would be a better method? I address the reason for this pull in issue 71.
https://github.com/rstudio/leaflet/issues/71
Thanks for making such a great piece of software. Sorry if I'm messing with your code flow.


yihui:
I don't know why you choose to merge into the `feature/better-defaults` branch. You should choose the master branch instead.

I think this NA issue is a general one (not specific to `addCircles()`), so we have to think about other `add*()` functions as well.

hansthompson:
Please forgive me.  I thought it fit under "better defaults" but after I wasn't sure and after a few minutes of thinking about it I just hit the button.

hansthompson:
Should I make a pull request to the master and close this one down?

yihui:
If you want, yes. But as I said, `addCircles()` does not seem to be the only thing that we need to fix. There are several other functions in this package that should consider NAs in lat/lng as well, and it will be nice to fix them all.

There are two possible ways to fix this issue: one is on the R side (exclude NA's from lat/lng), and the other is on the JavaScript side. I do not quite prefer the first way, since it may screw up other arguments such as `layerId`, e.g. the `layerId` vector will not match the lat/lng vectors after you exclude NAs from lat/lng.

@jcheng5 Any thoughts?

jcheng5:
I think it's only `addCircles`, `addMarkers`, and `addCircleMarkers` that are susceptible to NA. If you have, for example, polygon or rectangle data that includes NA points, that feels more like an error to me.

Out of curiosity, what kind of data do you have that you are ending up with NA lat/long values in the first place? Or is it NA radius (which would make more sense to me)?

hansthompson:
I'm doing an analysis for our bus tracking website, http://bus.codeforanchorage.org/. We store requests for real time bus data in gps, addresses, and bus stop ids. We get mostly gps requests which come out of our database a string. The script converts the requests that it can back to gps and then plots it. 

I could remove the NAs but do you think this could be categorized unexpected behaviour? I really like how ggplot handles NAs by putting them as a warning. Would that be a better implementation?

jcheng5:
I ran into this myself last week (real data with missing NA), it was quite annoying and I think you're right that we should warn. How about warning on the R side but actually doing the omitting on the JS side? I think that gives us best of both worlds.

masip85:
Yes,quite problematic issue. Every data after NA from your DF or DT is ommited and you can only see a portion of it. If you have a large dataset maybe the difference isn't noticed.
And rest of layers after NA's layer , doesn't plot anything.

bhaskarvk:
Closing as I'm tracking this under #304

PR #75, opened 2015-05-06
rename controlId to layerId for consistency with other add*/remove*/clear* methods

yihui:
cc @mplourde @jcheng5

### DIFF
--- a/R/controls.R
+++ b/R/controls.R
@@ -2,14 +2,13 @@
 #'   generated with Shiny/htmltools tags
 #' @param position position of control: 'topleft', 'topright', 'bottomleft', or
 #'   'bottomright'
-#' @param controlId the id of this control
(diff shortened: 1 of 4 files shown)


mplourde:
@yihui, this is no problem for me.

I chose `controlId`, because leaflet distinguishes between controls and layers (cf., [leaflet docs](http://leafletjs.com/reference.html#map-addlayer)). This means controls must be managed with something other than `LayerStore` in **inst/htmlwidgets/leaflet.js**. My initial thought was that `layerId` might be misleading.

Thanks.

yihui:
Thanks! I see your point. Since the docs of these functions are put in a single .Rd, I'd prefer not having two arguments that are basically the same thing (the tons of arguments on the help page `?addTiles` look a little overwhelming already :).

PR #81, opened 2015-05-07
Deal with single-point polygons/polylines

jcheng5:
This case breaks without this fix:

```r
library(maps)
leaflet(map(plot=FALSE)) %>% addPolylines()
```

It's because we expect arrays where single values are instead.


jcheng5:
@yihui Please review post-merge. Sorry, I'm on a roll with the docs and this is stopping me from being able to accurately preview one of my examples.

yihui:
That looks good to me. Thanks!

jcheng5:
@yihui Well, that caused an even worse bug in polygons. :) I guess I deserved that.

PR #82, opened 2015-05-08
Allow maps and lists to be used as data objects

jcheng5:


yihui:
Where does the `map` class come from?

jcheng5:
From the maps package

PR #83, opened 2015-05-08
Add topojson support via omnivore plugin

jcheng5:


yihui:
Do you need a method `methods.addTopoJSON()` in inst/htmlwidgets/leaflet.js accordingly?

jcheng5:
I've included it in a separate file, inst/htmlwidgets/plugins/leaflet-omnivore-plugin.js

I just realized I commented out the mouse handling and never went back and fixed it. Let's leave this open until I have a chance to do that.

yihui:
OK I see.

jcheng5:
Hey @yihui, I think this is ready now.

yihui:
LGTM. Is it worth squashing the commits into one? (since there are some files added and removed) I'm fine with merging it as is.

PR #85, opened 2015-05-11
Use CIE-Lab color interpolation instead of RGB

jcheng5:
Using Lab for color interpolation instead of RGB is recommended by Hadley and http://howaboutanorange.com/blog/2011/08/10/color_interpolation/

Unfortunately, `colorRamp(space = 'Lab')` is much, much slower than `colorRamp(space = 'rgb')`, but I have a fast implementation in rasterfaster that we can switch to once we are ready to take that dependency.

### DIFF
--- a/R/colors.R
+++ b/R/colors.R
@@ -258,7 +258,7 @@ NULL
 
 
 safePaletteFunc = function(pal, na.color) {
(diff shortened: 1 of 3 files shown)


yihui:
That sounds good to me.

PR #86, opened 2015-05-12
Invalidate map size when widget is resized

jcheng5:
Without this, then resizes that occur for reasons other
than the overall window being resized will not allow
tile layers to be rendered properly.

### DIFF
--- a/inst/htmlwidgets/leaflet.js
+++ b/inst/htmlwidgets/leaflet.js
@@ -821,8 +821,8 @@ var dataframe = (function() {
 
       setTimeout(function() { updateBounds(map); }, 1);
     },
-    resize: function(el, width, height, data) {
-
+    resize: function(el, width, height, map) {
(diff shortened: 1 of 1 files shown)


jcheng5:
Two examples:
https://gist.github.com/jcheng5/aff6ef5dbb76bb140f30

The "Shiny map" slide should render fine even without this fix; the others do need the fix.

To repro the problem, make sure when the browser loads the slide deck, that it's starting at slide 1.

yihui:
Hmm, I thought this would fix Tareef's ioslides problem. I tested it and it did not seem to.

jcheng5:
There was a related fix that was necessary in htmlwidgets. Installing both should fix it, lmk if not.

yihui:
I just installed htmlwidgets and indeed the issue has been fixed. Thanks!

Issue #87, opened 2015-05-12
Initial legend remains on map after Shiny input change


pssguy:
Sorry not sure if you view new comments  to #32 as it is closed

Thanks very much for this (addLegend). 

I want to use it in shiny with the user being able to change the number of bins via a numericInput.

However, the existing legend is not removed

A  [Work In Progress](https://mytinyshinys.shinyapps.io/blog/blog.Rmd)

pssguy:
Thanks for quick fix

PR #88, opened 2015-05-13
Refactor legend to use methods, controls

jcheng5:
This fixes the issue with legends not disappearing on re-render
of a leafletOutput, plus, lets legends be added/replaced/removed
using leafletProxy().


jcheng5:
Fixes issue #87

yihui:
That's great. Thanks! Do you mind using `layerId` instead of `controlId`?

jcheng5:
Oh, yeah, forgot we had switched that.

PR #90, opened 2015-05-13
Delay rendering of map until size != 0

jcheng5:
Leaflet does not behave well when you set up a bunch of layers
when the map is not visible (width/height == 0). Popups get
misaligned relative to their owning markers, and the fitBounds
calculations are off. Therefore we wait until the map is actually
showing to render the value (we rely on the resize() callback
being invoked at the appropriate time).


yihui:
I think it is okay to merge this one.

PR #95, opened 2015-05-19
Tweak docs

yihui:
Cc @jcheng5

### DIFF
--- a/docs/compile.R
+++ b/docs/compile.R
@@ -4,7 +4,7 @@ knitr::opts_chunk$set(out.width = '100%')
 library(leaflet)
 f = rmarkdown::render(commandArgs(TRUE))
 # remove version numbers in HTML
-r = '-\\d+[.]\\d+[.]\\d+$'
+r = '-\\d+[.]\\d+([.]\\d+)?$'
 v1 = rev(list.files('libs', r, full.names = TRUE))
 v2 = gsub(r, '', v1)
 x = readLines(f)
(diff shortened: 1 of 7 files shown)


docs/compile.R
```diff
@@ -1,9 +1,10 @@
 set.seed(0102)
 options(htmlwidgets.TOJSON_ARGS = list(pretty = TRUE))
+knitr::opts_chunk$set(out.width = '100%')
 library(leaflet)
 f = rmarkdown::render(commandArgs(TRUE))
 # remove version numbers in HTML
-r = '-\\d+[.]\\d+[.]\\d+$'
+r = '-\\d+[.]\\d+([.]\\d+)?$'
```
jcheng5: I don't actually know what this logic is for--do you?

yihui: The logic is like this:

- list all dirs under `./libs` of the form `lib-name-x.y(.z)?` (e.g. `bootstrap-3.1.0`, `leaflet-binding-0.0.15`, `htmlwidgets-0.4`)
- grep these dir names in the generated *.html
- replace `lib-name-x.y.z` with `lib-name`
- if there are two libs that only differ in the version number (e.g. `leaflet-binding-0.0.14` and `leaflet-binding-0.0.15`), remove the lower version

The purpose of this is to avoid lots of spurious changes when rebuilding the website and committing the changes, e.g. when htmlwidgets is upgraded from 0.3.3 to 0.4, I will only see this change (https://github.com/rstudio/DT/commit/56d23f46b70be4717c02d570163faab57c55ae7b) instead of a dir `htmlwidgets-0.3.3` being removed and a new dir `htmlwidgets-0.4` being added, in which case it is hard to know what really was changed.

PR #98, opened 2015-05-19
Remove spurious unbindAll calls

### LINKED ISSUE #97: Two updates of legend via inputmap_marker_click$id causes updates to stop
The first update works, but stops thereafter.  Also seen in other, more complex code, but here's a simple example.

library(shiny)
library(leaflet)

ui <- fluidPage(leafletOutput("map1"))

   values <- reactiveValues(click = "Click Markers")
   
   observeEvent(input$map1_marker_click$id,{
       values$click <- input$map1_marker_click$id
   })
   
      leaflet() %>% 
         addCircleMarkers(lng = runif(10), lat = runif(10), layerId = paste0("marker", 1:10)) %>%
         addLegend(position = "bottomright",
            colors = "blue",
            title = "Two clicks and no update of legend", 
            labels = values$click
         )
    })
    
}

shinyApp(ui, server)

jcheng5:
When legends are present, these resolve to Shiny.unbindAll(undefined)
since legends don't have _div fields.

It turns out that controls with _div fields already call Shiny.unbindAll
so there's no need to do it again here anyway.

Fixes #97 "Two updates of legend via inputmap_marker_click$id causes
updates to stop."

### DIFF
--- a/inst/htmlwidgets/leaflet.js
+++ b/inst/htmlwidgets/leaflet.js
@@ -229,14 +229,12 @@ var dataframe = (function() {
   ControlStore.prototype.clear = function() {
     for (var i = 0; i < this._controlsNoId.length; i++) {
       var control = this._controlsNoId[i];
-      Shiny.unbindAll(control._div);
       this._map.removeControl(control);
     };
     this._controlsNoId = [];
 
     for (var key in this._controlsById) {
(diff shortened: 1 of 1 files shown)


jcheng5:
See repro case in #97

PR #102, opened 2015-05-20
Fix #101: leafletProxy() function calls can get applied out-of-order

### LINKED ISSUE #101: leafletProxy() function calls can get applied out-of-order
```r
leafletProxy("map") %>% clearShapes() %>% addCircles()
```

I'm seeing clearShapes called after addCircles. This is because these function calls get individually deferred using `session$onFlushed` and it turns out those are not guaranteed to be called in the order they were registered.

Ideally this would be fixed in Shiny but we just released to CRAN this week. Another way to go would be to change leafletProxy from a list to a reference class or R6 class, that would at least let us ensure ordering WITHIN a single leafletProxy object (still sucky though).

jcheng5:
cc @wch

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -140,6 +140,23 @@ leafletProxy <- function(mapId, session = shiny::getDefaultReactiveDomain(),
   )
 }
 
+# Shiny versions <= 0.12.0.9001 can't guarantee that onFlushed
+# callbacks are called in the order they were registered. Rather
+# than wait for this to be fixed in Shiny and released to CRAN,
+# work around this for older versions by maintaining our own
+# queue of work items. The names in this environment are session
+# tokens, and the values are lists of invokeRemote msg objects.
+# During the course of execution, leafletProxy() should cause
+# deferred messages to be appended to the appropriate value in
+# sessionFlushQueue. It's the responsibility of invokeRemote to
+# ensure that the sessionFlushQueue values are properly reaped
+# as soon as possible, to prevent session objects from being
+# leaked.
+#
+# When Shiny >0.12.0 goes to CRAN, we should update our version
+# dependency and remove this entire mechanism.
+sessionFlushQueue = new.env(parent = emptyenv())
+
 invokeRemote = function(map, method, args = list()) {
   if (!inherits(map, "leaflet_proxy"))
     stop("Invalid map parameter; map proxy object was expected")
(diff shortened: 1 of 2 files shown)


R/utils.R
```diff
@@ -155,12 +172,45 @@ invokeRemote = function(map, method, args = list()) {
     )
   )
 
+  sess <- map$session
   if (map$deferUntilFlush) {
-    map$session$onFlushed(function() {
-      map$session$sendCustomMessage("leaflet-calls", msg)
-    }, once = TRUE)
+    if (compareVersion(as.character(packageVersion("shiny")), "0.12.0.9001") < 0) {
```
wch: You may know this already, but you can just do `packageVersion("shiny") < "0.12.0.9001"`.

R/utils.R
```diff
@@ -155,12 +172,45 @@ invokeRemote = function(map, method, args = list()) {
...
+
+      if (!exists(sess$token, sessionFlushQueue)) {
+        # If the current session doesn't have an entry in the sessionFlushQueue,
+        # initialize it with a blank list.
+        assign(sess$token, list(), sessionFlushQueue)
+
+        # If the session ends before the next onFlushed call, remove the entry
+        # for this session from the sessionFlushQueue.
+        endedUnreg <- sess$onSessionEnded(function() {
+          rm(list = sess$token, pos = sessionFlushQueue)
```
wch: Better to use `envir` instead of `pos` here.

R/utils.R
```diff
@@ -155,12 +172,45 @@ invokeRemote = function(map, method, args = list()) {
...
-      map$session$sendCustomMessage("leaflet-calls", msg)
-    }, once = TRUE)
+    if (compareVersion(as.character(packageVersion("shiny")), "0.12.0.9001") < 0) {
+
+      # See comment on sessionFlushQueue.
+
+      if (!exists(sess$token, sessionFlushQueue)) {
+        # If the current session doesn't have an entry in the sessionFlushQueue,
+        # initialize it with a blank list.
+        assign(sess$token, list(), sessionFlushQueue)
```
wch: I just want to make sure you're aware that it's much more efficient to do `sessionFlushQueue[[sess$token]] <- list()`.

In general, functions like get/assign/exists are much slower than `[[`. See this thread if you want to know more: 
  http://r.789695.n4.nabble.com/we-need-an-exists-get-hybrid-td4700333.html

wch:
Looks good to me.

jcheng5:
Thanks for the review. Once the Travis tests pass, I'll merge.

jcheng5:
The test I added is failing on Travis, but works for me locally. UGH

PR #103, opened 2015-05-21
Leaflet.Sleep

yihui:
See https://github.com/CliffCloud/Leaflet.Sleep for more info.

I think it makes sense to include this plugin by default. Two issues:

1. Currently I set `hoverToWake` to `false`, so the only way to enable zooming is to click on the map; I feel hovering to wake the map may also be undesirable when one scrolls up/down the page.
2. There is a little opacity when the map sleeps. I'm not sure if that is desirable. It is hard-coded in the JS source of the plugin.

### DIFF
--- a/inst/htmlwidgets/leaflet.js
+++ b/inst/htmlwidgets/leaflet.js
@@ -717,7 +717,9 @@ var dataframe = (function() {
       // is no way for htmlwidgets to pass initial params to initialize()
       var map = L.map(el, {
         center: [51.505, -0.09],
-        zoom: 13
+        zoom: 13,
+        sleepNote: false,
+        hoverToWake: false
       });
 
       // Store some state in the map object
--- a/inst/htmlwidgets/leaflet.yaml
+++ b/inst/htmlwidgets/leaflet.yaml
@@ -12,3 +12,7 @@ dependencies:
     version: 1.0.0
     src: "htmlwidgets/lib/leafletfix"
     stylesheet: leafletfix.css
+  - name: Leaflet.Sleep
+    version: 2015.02.25
+    src: "htmlwidgets/plugins/Leaflet.Sleep"
+    script: Leaflet.Sleep.js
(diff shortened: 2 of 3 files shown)


jcheng5:
I strongly dislike both `hoverToWake=false` and the opacity change. Even with `hoverToWake=true` and the opacity gone, I'm not sure this plugin gives a good enough experience for us to ship. I feel like if I put my mouse over the map and scroll, it shouldn't ignore me, even for 750 milliseconds.

yihui:
Fair enough. I'll see if it is possible to detect if a mouse over event is "active" or "passive" (caused by scrolling).

yihui:
I read a few SO posts and this Github issue https://github.com/Leaflet/Leaflet/issues/2076 (Good news is the Leaflet author thought it was a good idea to prevent zooming when scrolling was from outside) I also tested this https://github.com/mapbox/deadsea but it doesn't work as expected. It seems the element is always unlocked by the mousemove event (https://github.com/mapbox/deadsea/blob/mb-pages/deadsea.js#L34).

It is super easy to manually activate the zooming mode by clicking on the map (e.g. http://gis.stackexchange.com/q/111887/52501) but apparently you don't like it. I feel it is very tricky to solve this problem by checking where scrolling came from: if you scroll down the page, and happen to have stopped on a map (just because you have reached the end of the touch pad or the mouse wheel), then scroll again, will that mean zooming or continue scrolling?

jcheng5:
I haven't tried the deadsea one, but what's wrong with the element being unlocked by mousemove? That was the approach I was thinking of.

yihui:
It is totally fine and desired for the element being unlocked by mousemove. The problem is, mousemove is also triggered by scrolling, even if you don't actively move the mouse. The mouse is considered to be moving as long as its position is changing relative to the element, so if the map is moving upwards/downwards and the mouse does not move, the mousemove event is still triggered. That is the challenge.

PR #104, opened 2015-05-21
closes #57: add the setMaxBounds() method

### LINKED ISSUE #57: setting maxBounds for leaflet()
wondering if we can set maxBounds for the map so the user cannot zoomout to space...thank you.

yihui:


R/methods.R
```diff
@@ -52,6 +52,24 @@ fitBounds = function(map, lng1, lat1, lng2, lat2) {
...
+#' @describeIn map-methods Restricts the map view to the given bounds
+#' @export
+setMaxBounds = function(map, lng1, lat1, lng2, lat2) {
+  bounds = list(lat1, lng1, lat2, lng2)
+
+  dispatch(
+    map,
+    "setMaxBounds",
+    leaflet = {
+      map$x$setMaxBounds = bounds
```
jcheng5: Why set this on x instead of just doing invokeMethod?

yihui: Good point. Just changed it to `invokeMethod()`.

PR #105, opened 2015-05-22
MarkerCluster

yihui:
This is a first attempt at integrating the MarkerCluster plugin (I'm posting this PR to see if there are better alternative approaches). Since the usage of this plugin is based on markers, I just added two more arguments to `addMarkers()`. To enable the plugin, specify `clusterOptions` as a list, e.g.

```r
devtools::install_github('rstudio/leaflet@feature/MarkerCluster')

library(leaflet)
leaflet() %>% addTiles() %>% addMarkers(
  data = quakes, popup = ~ sprintf('magnitude = %s', mag), clusterOption = list()
)
```

screenshot


inst/htmlwidgets/leaflet.js
```diff
@@ -401,19 +402,32 @@ var dataframe = (function() {
 
     if (icon) icondf.effectiveLength = df.nrow();
 
+    var clusterGroup, cluster = clusterOptions !== null;
+    if (cluster) {
+      clusterGroup = L.markerClusterGroup(clusterOptions);
```
jcheng5: If a cluster with this `clusterId` already exists, shouldn't we add to it?

jcheng5: Actually I'm a little unclear on what the ownership of a marker should be.

```r
shinyServer(function(input, output, session) {
  output$map <- renderLeaflet({
    leaflet() %>% addMarker(0, 0, layerId = "marker1")
  })
  observe({
    clusterOptions <- if (input$cluster) list() else NULL
    leafletProxy() %>% addMarker(0, 0, layerId = "marker1", clusterGroup = clusterOptions)
  })
})
```

I'd expect each time the observe executes, it would replace the marker, but I don't think this is going to happen...?

yihui: Good points. Let me think more about it. Thanks!

jcheng5:
TODO:

- [ ] When adding a marker to a marker cluster, be sure to remove any marker on that same cluster with the same `layerId`
- [ ] `removeMarkerFromCluster(clusterId, layerId)`
- [ ] `removeMarkerCluster(clusterId)`
- [ ] `clearMarkerClusters()`

jcheng5:
We also need a convenience function for creating cluster options.

jcheng5:
Looks like tests need updating, I can help you with that if it's not clear.

Can you also have mouse events back to Shiny include the cluster ID? Since layer IDs are namespaced by cluster ID I think it's important that the latter be included.

yihui:
Tests updated. Conflicts resolved. For the cluster ID, do you want it to be part of `input$map_marker_click` (e.g. `list(id = 'markerId', clusterId = 'clusterId', lat = , lng = )`), or separate like `input$map_cluster_click`? Currently I did the latter since it is relatively straightforward.

inst/htmlwidgets/leaflet.js
```diff
@@ -482,19 +517,39 @@ var dataframe = (function() {
...
     if (icon) icondf.effectiveLength = df.nrow();
 
+    var clusterGroup = this.markerclusters.get(clusterId),
+        cluster = clusterOptions !== null;
+    if (cluster && !clusterGroup) {
+      clusterGroup = L.markerClusterGroup(clusterOptions);
+      clusterGroup.clusterLayerStore = new ClusterLayerStore(clusterGroup);
+    }
+    (function() {
+      clusterGroup.on('clusterclick', mouseHandler(this.id, clusterId, 'cluster_click'), this);
```
jcheng5: This will throw an error if `addMarkers` is called without clustering.

jcheng5:
> For the cluster ID, do you want it to be part of `input$map_marker_click` (e.g. `list(id = 'markerId', clusterId = 'clusterId', lat = , lng = )`), or separate like `input$map_cluster_click`?

I was thinking the former. Wherever the event is surfaced, it needs to have both the layer ID and the cluster ID.

yihui:
Okay, I think everything is done now.

Issue #73, opened 2015-05-04
FR: support ImageOverlay


name-name:
Hi is it possible or planned to support leaflet's ImageOverlay: http://leafletjs.com/reference.html#imageoverlay

Specifically I'd like to add a transparent plot generated by R to the map (some splines through the markers), which should be zoomed/panned etc when the map is.

jcheng5:
I'm planning on implementing that this week. Do you have a code sample for how you're generating the image?

name-name:
Well, creating a spline curve in 2D with xspline for a fixed set of points is explained here:
http://stackoverflow.com/questions/28093925/calculate-a-2d-spline-curve-in-r

However, I'm not quite sure if this works at all on a zoomable map:
1) I probably need to project coordinates to the 2D points then create the plot and calculate the image bounds as lat/long (i.e. reverse projection)
2) not sure if having a fixed size image with given "imageBounds" (in lat/long) is enough when we zoom - but looking at a VERY simple (right-angled) triangle on a map: zooming doesn't change the angles!

Tested with this:

        ## app.R ##
        library(shiny)
        library(shinydashboard)
        library(leaflet)
        
        lng = c(0,0,10,0)
        lat = c(0,10,0,0)
        triangle = data.frame(lat,lng)
        
        
        runApp(list(
          ui = dashboardPage(
            dashboardHeader(title = ""),
            dashboardSidebar(disable=T),
            dashboardBody(
                leafletOutput("map", height = 600)
            )
          ),
          server = function(input, output) {
            
            output$map <- renderLeaflet({
              
              map <- leaflet() %>% addTiles()
              
              map %>% addPolylines(data=triangle, fillOpacity = 1.0,
                                   lat = ~lat, lng = ~lng)  # :-( 
            })
          }
        ))

jcheng5:
> I probably need to project coordinates to the 2D points then create the plot and calculate the image bounds as lat/long (i.e. reverse projection)

This kind of thing is why I wanted a code sample--I wanted to know whether you are generating a projected or unprojected image.

And pardon my ignorance, but why would you expect the angles to change as you zoom?

name-name:
I added the leaflet link in my comment above. Either IProjection is available in the leaflet package (maybe another FR^^) or I code it myself in R - shouldn't be that hard:
https://github.com/Leaflet/Leaflet/tree/master/src/geo/projection

[...]

PR #110, opened 2015-05-28
Raster image support

jcheng5:
This branch introduces `addRasterImage`, `removeImage`, and `clearImages` functions. These functions can be used to incorporate small-ish RasterLayer objects (from the `raster` package) into maps, by (optionally) projecting them, coloring them, PNG and base64 encoding them, and finally embedding them directly into the widget data.

There's some complexity in the JS file, it turns out that Leaflet's ImageOverlay feature didn't work reliably for me and didn't support nearest-neighbor interpolation.

Incidentally, there's now an `alpha` parameter to the color functions. If `TRUE` then alpha interpolation will be performed and `#RRGGBBAA` strings will be returned.

**Note that this PR introduces a dependency on jcheng5/rasterfaster.** This may change to hadley/scales if he accepts https://github.com/hadley/scales/pull/57, though rasterfaster will come back when we get around to making addRasterTiles work properly.


jcheng5:
Ack, just realized I am missing some options.

- [ ] Min/max zoom levels
- [ ] detectRetina
- [ ] reuseTiles

Also, note that this feature doesn't work in older IE versions as it relies on canvas.

jcheng5:
Tested on Chrome and Firefox on Linux, and IE10 on Windows 8.

jcheng5:
Fixes #73

jcheng5:
Change rasterfaster dependency to scales

jcheng5:
@yihui I can't remember, did I say I was going to fix something else besides moving from `rasterfaster` dependency to `scales`?

yihui:
I don't remember, either.

PR #116, opened 2015-06-09
Add groups and addLayersControl features

jcheng5:
To install the branch:

```r
devtools::install_github("rstudio/leaflet@joe/feature/layers-control")
```

This adds a few new parameters and functions. First of all, almost every function that adds a layer (`addTiles`, `addMarkers`, etc.) now takes a `group` parameter; this must be either `NULL` or a single-element character vector. The group is useful for two things:

1. You can create one to many layers with the same group, and then remove them all at once with `clearGroup(groupname)` (this is mostly useful when used in conjunction with Shiny and `leafletProxy()`, see [the guide](http://rstudio.github.io/leaflet/shiny.html)).
2. You use groups to identify which layers you want in the Layers Control.

An example (see [rendered output](http://rpubs.com/jcheng/leaflet-layers-example)):

```r
leaflet() %>%
  # Base maps
  addTiles(group = "OpenStreetMap") %>%
  addProviderTiles("Stamen.Toner", group = "Stamen Toner") %>%
  # Overlays
  addMarkers(runif(20, -90, -80), runif(20, 40, 50), group = "Markers") %>%
  addCircleMarkers(runif(20, -90, -80), runif(20, 40, 50), group = "Circle Markers") %>%
  # The layers control
  addLayersControl(
    baseGroups = c("OpenStreetMap", "Stamen Toner"),
    overlayGroups = c("Markers", "Circle Markers")
  )
```

Open issues/TODO:

- [x] Not currently possible to specify which basemaps/overlays you want visible by default
- [x] Not possible to remove layers control, and calling `addLayersControl` more than once will result in multiple layers controls
- [x] Should send Shiny inputs about what basemaps/overlays the user has selected
- [ ] <s>It might be nice to let the user specify identifier-like strings for group names (`"stamen_toner"`) while still having human readable labels in the layers control (`"Toner (by Stamen)"`).</s>
- [x] All unit tests need to be updated to accommodate the new `group` argument
- [ ] Allow controls/legends to be attached to groups


cawthm:
so addLayersControl will anticipate only two layers: baseGroup and overlayGroup?

while unusual, I could think of situations where you'd want more layer depth.

you could specify the arguments like: 

...%>% addLayersControl(groupOrder = c("group_layer_bottom", c("group_layer_middle_1", "group_layer_middle_2",...,"group_layer_middle_n"), "group_layer_top"))

Just a thought.  Thanks for adding this.

jcheng5:
@cawthm It's two *sets* of layers (actually groups). You can supply any number of groups to baseGroups, but the user will have to select ONE of them to be displayed. You can supply any number of groups to overlayGroups, and the user can select whether each of them should be visible or not.

Each group can be made up of many layers. So you wouldn't have `"group_layer_middle_1"` through `"group_layer_middle_n"`; instead all of those layers would share the group name `"group_layer_middle"`.

cawthm:
got it.  makes sense.  thank you.

R/layers.R
```diff
@@ -715,3 +729,61 @@ removeGeoJSON = function(map, layerId) {
...
+#'   most commonly used for mostly-opaque tile layers.
+#' @param overlayGroups character vector where each element is the name of a
+#'   group. The user can turn each overlay group on or off independently.
+#' @param position position of control: 'topleft', 'topright', 'bottomleft', or
+#'   'bottomright'
+#' @param options a list of additional options, intended to be provided by
+#'   a call to \code{layersControlOptions}
+#'
+#' @examples
+#' \donttest{
```
yihui: CRAN may ask us to explain why `\donttest{}`. I guess it is fine not to use `\donttest{}` here, since I don't see how it can be problematic. Or we add examples to the website, and provide a link to the web page here.

PR #122, opened 2015-06-17
closes #121: the default title from deparse(values) may be bad

### LINKED ISSUE #121: Bad default title for legend
Check out what this code produces:

```r
library(raster)

r <- raster("nc/oisst-sst.nc")
pal <- colorNumeric(c("#0C2C84", "#41B6C4", "#FFFFCC"), values(r),
  na.color = "transparent")

leaflet() %>% addTiles() %>%
  addRasterImage(r, colors = pal, opacity = 0.8) %>%
  addLegend(pal = pal, values = values(r))
```

Results:
http://rpubs.com/jcheng/bad-legend-title

I'd be totally OK with legends having no title by default, what do you think @yihui?

yihui:
when values is a formula, I think it still makes sense to deparse() the RHS of the formula by default: https://github.com/rstudio/leaflet/blob/0bcdd2d52500ad17043f3973cbd3856df9a09179/R/legend.R#L68

### DIFF
--- a/R/legend.R
+++ b/R/legend.R
@@ -53,7 +53,7 @@
 addLegend = function(
   map, position = c('topright', 'bottomright', 'bottomleft', 'topleft'),
   pal, values, na.label = 'NA', bins = 7, colors, opacity = 0.5, labels,
-  labFormat = labelFormat(), title = deparse(substitute(values)),
+  labFormat = labelFormat(), title = NULL,
   layerId = NULL
 ) {
   position = match.arg(position)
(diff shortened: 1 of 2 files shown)


jcheng5:
OK. Because the test is `missing(title)`, if you are using a formula and don't want a title you can still pass `title = NULL`, is that right?

If so, LGTM--feel free to merge once Travis succeeds.

yihui:
You are right. If you don't want the title, you can use `title = NULL` explicitly.

PR #123, opened 2015-06-17
Interfere less with page scrolling (e.g. in Rmd)

jcheng5:
This is an alternative to #103


jcheng5:
Demo: http://rpubs.com/jcheng/leaflet-scroll-test

yihui:
That is really nice!! :+1:

PR #126, opened 2015-06-19
Allow legend CSS class to be customized

jcheng5:
Also changed addControl's classes argument to be consistent with className


jcheng5:
@yihui I think this is the last code change.

yihui:
Perfect. Thanks!

PR #131, opened 2015-06-23
SVG numeric legend

jcheng5:
@yihui this works and seems a better long-term approach. But I want to sleep on it and ideally discuss with you tomorrow before deciding whether it's worth including such a large code change for the initial release.

### DIFF
--- a/inst/htmlwidgets/lib/leafletfix/leafletfix.css
+++ b/inst/htmlwidgets/lib/leafletfix/leafletfix.css
@@ -17,10 +17,23 @@ img.leaflet-tile {
     line-height: 18px;
     color: #555;
 }
+.legend svg text {
+    fill: #555;
+}
+.legend svg line {
+    stroke: #555;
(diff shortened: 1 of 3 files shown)


inst/htmlwidgets/lib/leafletfix/leafletfix.css
```diff
@@ -24,3 +30,10 @@ img.leaflet-tile {
     margin-right: 8px;
     opacity: 0.7;
 }
+.legend table {
+    border-spacing: 2px;
+    border-collapse: separate;
+}
+.legend td {
+    padding: 1px;
+}
```
yihui: I guess `.legend table` and `.legend td` are no longer used?

jcheng5: Oh, I just assumed the non-numeric legends were using them. I'll delete them.

PR #163, opened 2015-08-12
Added removeFeatureGeoJSON, addFeatureGeoJSON, and styleFeatureGeoJSON

ideamotor:
This PR adds a simple version of the aforementioned functions for GeoJSON. For `removeFeatureGeoJSON`, you provide a featureId and layerId (you must have an id in in the GeoJSON, such as geojson$features[[1]]$id which `leaflet-shiny` returns as featureId on events. `addFeatureGeoJSON` adds a feature to an existing GeoJSON. Finally, `setStyleGeoJSON` also currently requires a style in JSON.

There is also a new example, shiny_interactions.R that shows many of the use cases of these functions. You can remove and add back features and then restyle the remaining features, and features are highlighted with mouse hover events. Please note that in a real-use case scenario, the data upon which the coloring is based would not be static. The speed could be improved by setting many styles at once instead of performing a loop for each feature. In addition, the highlighting functionality could be handled on the client side without passing events to R.


jcheng5:
@yihui can you do an initial review of this please? I'll also review after JSM (maybe next week).

yihui:
I just reviewed the code and it looks pretty good to me. I'll test the examples later. Thanks @cholcomb!

cholcomb:
I noticed line 5 of shiny_interactions.R says `library(rescale)` instead of `library(scales)`.... Also we probably want to rename `setStyleGeoJSON` to `styleFeatureGeoJSON` because this function is just for styling one feature. I went ahead and committed those changes.

R/layers.R
```diff
@@ -943,6 +943,53 @@ addGeoJSON = function(map, geojson, layerId = NULL, group = NULL,
...
+  if(class(layerId) != "character" || length(layerId) != 1) {
+    stop("Missing layerId string")
+  }
+  if(class(featureId) != "character" || length(featureId) != 1) {
+    stop("Missing featureId string")
+  }
+  invokeMethod(map, getMapData(map), 'removeFeatureGeoJSON', layerId, featureId)
+}
+
+#' @param data in GeoJSON
```
yihui: I think you need to get rid of this line.

R/layers.R
```diff
@@ -943,6 +943,53 @@ addGeoJSON = function(map, geojson, layerId = NULL, group = NULL,
...
+#' @export
+styleFeatureGeoJSON = function(map, layerId = NULL, featureId = NULL, style = NULL) {
+  if(class(layerId) != "character" || length(layerId) != 1) {
+    stop("Missing layerId string")
+  }
+  if(class(featureId) != "character" || length(featureId) != 1) {
+    stop("Missing featureId string")
+  }
+  if(class(style)=="list") {
+    styleJSON <- as.character(jsonlite::toJSON(style))
```
yihui: Why manually encode the list here and `JSON.parse()` later in JS? This seems to be unnecessary to me. I'd remove the check of `class(style)` here, accept whatever the user inputs, then in JS use `JSON.parse()` only if necessary (cf `methods.addGeoJSON()`).

R/layers.R
```diff
@@ -943,6 +943,53 @@ addGeoJSON = function(map, geojson, layerId = NULL, group = NULL,
   invokeMethod(map, getMapData(map), 'addGeoJSON', geojson, layerId, group, options)
 }
 
+#' @param style a JSON string
+#' @describeIn map-layers change style of GeoJSON feature
+#' @export
+styleFeatureGeoJSON = function(map, layerId = NULL, featureId = NULL, style = NULL) {
+  if(class(layerId) != "character" || length(layerId) != 1) {
```
yihui: `class(foo) == 'bar'` is not an appropriate way to test the class in R, since an object may have multiple classes. Use `inherits(foo, 'bar')` if the function `is.bar()` does not exist (in this case, just use `is.character()`).

inst/htmlwidgets/leaflet.js
```diff
@@ -856,6 +858,36 @@ var dataframe = (function() {
     });
   };
 
+		// TO DO: right now using .id but may let users choose, let users choose "<" or ">" or "=" (default)
+		// consider letting users specify listed style arguments ala addGeoJSON
+		// and create methods for many ids to style or remove with single loop, and many ids to many styles
```
yihui: Yes, it will be nice to vectorize over id or/and style.

(nitpicking: could you replace tabs with spaces here?)

inst/examples/shiny_interactions.R
```diff
@@ -0,0 +1,247 @@
...
+library(leaflet)
+library(rgdal)
+library(jsonlite)
+library(scales)
+googleLink <- function(lat,lng,zoom) {
+  sprintf("http://www.google.com/maps/place/%s,%s/@%s,%s,%sz/data=!3m1!1e3",
+          lat, lng, lat, lng, zoom)
+}
+geojson <- paste(readLines(system.file("examples/countries.json", package = "leaflet")), collapse = "\n") %>%
+  fromJSON(simplifyVector = FALSE)
```
yihui: I believe paste and readLines are not necessary since `fromJSON()` can read from a file.

inst/examples/shiny_interactions.R
```diff
@@ -0,0 +1,247 @@
...
+    leafletProxy("mymap") %>%
+      addFeatureGeoJSON(as.character(minify(bermudaTriangle)), layerId ='geojsonlayer') # can use a GeoJSON string (as.character)
+  })
+  observeEvent(input$removeBermuda, {
+    leafletProxy("mymap") %>%
+      removeFeatureGeoJSON(layerId ='geojsonlayer', featureId = "Bermuda Triangle")
+  })
+}
+shinyApp(ui, server)
+
```
yihui: The example is very nice :+1: The interaction (mouseover) is a bit slow, though. I didn't investigate this issue.

inst/htmlwidgets/leaflet.js
cholcomb: Yes it's slow. We really need a way to send over setStyle or resetStyle functions to the client, although I'm not sure yet how to do that and allow customization. Note: I don't believe resetStyle will work for when you effectively want to change back to the prior style instead of the originally loaded style (when you have dynamic data). Internally, on `leaflet.js` we have hard-coded:

```javascript
    function highlightFeature(e) {
    	var layer = e.target;
			layer.setStyle({weight: 3,color: "black"});
	if (!L.Browser.ie && !L.Browser.opera) {
		layer.bringToFront();};
    };
    function resetHighlight(e) {
      var layer = e.target;
      var originalColor = e.target.feature.properties.style.color;
      var originalWeight = e.target.feature.properties.style.weight;
      layer.setStyle({weight: originalWeight,color: originalColor});
    };
```

and inside the `onEachFeature` of `var gjlayer` of `methods.addGeoJSON` we include:

```javascript
    layer.on("mouseover", highlightFeature , this);
    layer.on("mouseout", resetHighlight, this);
```

The capability of assigning styles to events could probably be added to `addGeoJSON`. Then, we probably need some sort of function `styleGeoJSON` that lets you restyle an entire layer (many users won't actually have dynamic data), and a function `styleEventGeoJSON` that lets you restyle the event styles .

R/layers.R
```diff
@@ -943,6 +943,42 @@ addGeoJSON = function(map, geojson, layerId = NULL, group = NULL,
   invokeMethod(map, getMapData(map), 'addGeoJSON', geojson, layerId, group, options)
 }
 
+#' @param style a JSON string
```
yihui: or a list

inst/examples/shiny_interactions.R
cholcomb: Yes it's slow. We really need a way to send over setStyle or resetStyle functions to the client, although I'm not sure yet how to do that and allow customization. Note: I don't believe resetStyle will work for when you effectively want to change back to the prior style instead of the originally loaded style (when you have dynamic data). Internally, on `leaflet.js` we have hard-coded:

```javascript
    function highlightFeature(e) {
    	var layer = e.target;
			layer.setStyle({weight: 3,color: "black"});
	if (!L.Browser.ie && !L.Browser.opera) {
		layer.bringToFront();};
    };
    function resetHighlight(e) {
      var layer = e.target;
      var originalColor = e.target.feature.properties.style.color;
      var originalWeight = e.target.feature.properties.style.weight;
      layer.setStyle({weight: originalWeight,color: originalColor});
    };
```

and inside the `onEachFeature` of `var gjlayer` of `methods.addGeoJSON` we include:

```javascript
    layer.on("mouseover", highlightFeature , this);
    layer.on("mouseout", resetHighlight, this);
```

The capability of assigning styles to events could probably be added to `addGeoJSON`. Then, we probably need some sort of function `styleGeoJSON` that lets you restyle an entire layer (many users won't actually have dynamic data), and a function `styleEventGeoJSON` that lets you restyle the event styles.

ADDED: It looks like the changes you suggested sped it up considerably, thanks! Within the "styled events" we also need a way to send features to the top using `bringToFront` as in this example (if you look at the borders, some are thinner because they are below other features).

Despite being faster, it does still get suck with some highlighted features that you have mouseout'd on (it does not do this on the hard-coded JS version).

UPDATE: moved this comment to here

jcheng5:
Sorry guys, is this one ready to go? Were you waiting on me?

jafflerbach:
Hi @jcheng5 - this feature would be very useful for our OHI Shiny App any status update??

jcheng5:
Not at the moment, but we just hired a new software engineer and she'll likely be working on Leaflet first. So help is on the way :)

bhaskarvk:
I created a new PR #303 for this with the new way of doing JS stuff.

PR #171, opened 2015-08-24
add leaflet-measure control

timelyportfolio:
[`leaflet-measure`](https://github.com/ljagis/leaflet-measure) seemed like a very good and easy addition to `leaflet`.  It adds a control to measure a polygon that a user can draw onto a `leaflet` map.  I tried to model the pull on others that I have seen recently.  Please let me know if there are any consistencies.

Here is an example with screenshot.

```r
library(leaflet)

leaf <- leaflet() %>%
  addTiles() %>%
  # central park
  fitBounds(-73.9, 40.75, -73.95,40.8) %>%
  addMeasure()

leaf

# to remove the measure
leaf %>% removeMeasure()
```

image


inst/htmlwidgets/leaflet.js
```diff
@@ -1429,6 +1429,16 @@ var dataframe = (function() {
     this.layerManager.clearLayers("image");
   };
 
+  methods.addMeasure = function(options){
+    this.measureControl = L.control.measure(options)
+    this.measureControl.addTo(this);
+  };
+
+  methods.removeMeasure = function(layerId){
```
jcheng5: No need for `layerId` parameter

R/plugin-measure.R
```diff
@@ -0,0 +1,39 @@
...
+
+#' Add a measure control to the map.
+#'
+#' @param measureOptions \code{list} of options as described in \href{https://github.com/ljagis/leaflet-measure#control-options}{leaflet-measure}.
+#' @return modified map
+#' @examples
+#' leaflet() %>%
+#'   addMeasure()
+#' @export
+addMeasure <- function( map, measureOptions=NULL ) {
```
jcheng5: Instead of `measureOptions=NULL` it would be much better to have explicit parameters for the important options (which in this case, may be all of them). This is particularly important because it allows autocomplete features in R IDEs and GUIs to be an easy reminder of what options are available.

inst/htmlwidgets/leaflet.js
```diff
@@ -1429,6 +1429,16 @@ var dataframe = (function() {
     this.layerManager.clearLayers("image");
   };
 
+  methods.addMeasure = function(options){
+    this.measureControl = L.control.measure(options)
```
jcheng5: Should check if `this.measureControl` already exists, and if so, remove it.

jcheng5:
Other than the comments I made this looks great, thank you!

timelyportfolio:
@jcheng5, very good suggestions. Thanks for the review.  I think all changes have been made.

jcheng5:
Great! Last thing we need from you is a signed CLA, if you don't mind. https://www.rstudio.com/wp-content/uploads/2014/06/RStudioIndividualContributorAgreement.pdf If scanning isn't convenient, you can just email a cellphone pic of the signed agreement to joe@rstudio.com.

The CLA assigns copyright to us so we can relicense our package in the future if necessary, e.g. to a different version of the GPL or to a more liberal BSD/MIT style license. (Without having CLAs on file for all contributors, we'd have to try to wrangle a unanimous vote to make these kinds of changes.)

jcheng5:
I received the CLA, thank you.

PR #177, opened 2015-09-02
added Leaflet.draw support

gherardovarando:
In particular the function addDrawToolbar.


R/plugin-draw.R
```diff
@@ -0,0 +1,37 @@
+
+
+leafletDrawDependencies <- function() {
+  list(
+    htmltools::htmlDependency(
+      "Leaflet.draw",
+      "1.0.0",
```
jcheng5: This should be `"0.2.3"`, per the package.json file in Leaflet.draw.

R/plugin-draw.R
```diff
@@ -0,0 +1,37 @@
...
+#' @param map the map to add the toolbar to
+#' @return modified map object
+#' @export
+#' @examples
+#' map <- addTiles(leaflet())
+#' map <- addDrawToolbar(map)
+#' map
+addDrawToolbar = function(map) {
+ map$dependencies <- c(map$dependencies, leafletDrawDependencies())
+ map$drawToolbar<-T
```
jcheng5: What's this for?

inst/htmlwidgets/plugins/leaflet-draw-plugin/leaflet-draw-plugin.js
```diff
@@ -0,0 +1,19 @@
...
+   edit: {
+       featureGroup: drawnItems
+        }
+     });
+  this.addControl(drawControl);
+  this.on('draw:created', function (e) {
+			var type = e.layerType,
+				layer = e.layer;
+			if (type === 'marker') {
+				layer.bindPopup('A popup!');
```
jcheng5: Probably didn't mean to include this? :)

inst/htmlwidgets/plugins/leaflet-draw-plugin/leaflet-draw-plugin.js
```diff
@@ -0,0 +1,19 @@
+  LeafletWidget.methods.addDrawToolbar = function(){
```
jcheng5: Do you mind cleaning up the indentation in this file? Two spaces per level, please! :)

jcheng5:
Thanks for this contribution, it's a great start!

Before we can proceed we need a signed Contributors License Agreement from you. This is necessary to let us make changes to the licensing in the future if desired, e.g. if we want to move to a more liberal license or a future version of the GPL.

https://www.rstudio.com/wp-content/uploads/2014/06/RStudioIndividualContributorAgreement.pdf If scanning isn't convenient, you can just email a cellphone pic of the signed agreement to joe@rstudio.com.

I also would like some (ok, lots of) additional features. If you're not up for implementing them, one of our team can do it (but I'm not sure exactly when it'll happen, so any contributions are gratefully accepted).

* Make drawing options available
* `removeDrawToolbar` (or maybe instead of `add`/`remove` it should be `show`/`hide`?)
* Let the addDrawToolbar specify the `layerId` of drawn items
* `removeDrawing(map, layerId)` and `clearDrawings(map)`
* The drawn items need to be submitted to Shiny, using `Shiny.onInputChange(name, value)`

R/plugin-draw.R
gherardovarando: nothing, sorry.

inst/htmlwidgets/plugins/leaflet-draw-plugin/leaflet-draw-plugin.js
gherardovarando: no, thanks

gherardovarando:
- add/remove DrawToolbar 
- is possible to specify layerId of drawn itmes
- the items are submitted to shiny when created, not yet when edited or removed

jcheng5:
Thanks for the additional commits! I'm underwater getting ready for a conference next week, but I will review when I get back online.

gherardovarando:
when edit or delete action start/stop the event is passed to Shiny through logical input.
Example: edit start: input$layerID_editstart == TRUE and input$layerID_editstop == FALSE
                editstop:    input$layerID_editstart == FALSE and input$layerID_editstop == TRUE

cholcomb:
Hey guys, we did an initial test of this and like the direction, there are a couple issues however:
 * Drawing lines and then attempting to edit anything causes an error (it's looking for the fill option)
 * AKAIK, there isn't a input$layerID_drawnItems. I.e. something that captures what's on the map. input$layerID_create only keeps what is the most recently initially created polygon or set of polygons, while input$layerID_edit only keeps the most recently edited polygon. You would need to write some odd logic to capture this.
 * Similar to this I feel that there is a need for an featureId for each current (aka drawn Item), either arbitrary or set by the user. Right now you are limited to a singular layerId for all the drawn items.
 * Related to featureId's, the user should be able to change the fill, color, opacity, and weight of each drawn item as they are created (before) and when they are edited. Integration with shinyJS colourInput would be ideal, although as described later this could be left outside this package.

Generally, in keeping with the API style of shiny and leaflet, I feel that we need to be able to add elements to the leaflet proxy on the basis on events such as actionButton clicks. I.e., use functions like:
* leafletProxy("mymap") %>% createFeatureGeoJSON(layerId=X,featureId=Y,type="circle",style=list(fill="red",weight=3)
* leafletProxy("mymap") %>% editFeatureGeoJSON(layerId=X,featureId=Y)

Then you can use functions from rstudio/leaflet#163:
* leafletProxy("mymap") %>% styleFeatureGeoJSON(layerId=X,featureId=Y,style=list(color="red",weight=13))
* leafletProxy("mymap") %>% removeFeatureGeoJSON(layerId=X,featureId=Y)

This would let folks easily adapt drawing tools into their own UI and tightly control when specific functionality is available.

jcheng5:
@cholcomb I love the idea of "createFeatureGeoJSON" (though I would like it even better if there were separate functions for each shape type).

Triggering drawing operations is pretty easy to do programatically:

```javascript
var options = {shapeOptions: {fillColor: "red", weight: 13}};
var circle = new L.Draw.Circle(this, options);
circle.enable();
```

The trickiest part might be to catch the created item, which is accessible via the `draw:created` event of the map (not the L.Draw.Circle... which would be nicer in this case).

Some other examples of editing existing shapes is [here](https://github.com/Leaflet/Leaflet.draw/blob/master/examples/edithandlers.html).

As much as I like it, I think that it's probably a good idea to separate this functionality into a separate PR, as I hate to hang up the toolbar-style drawing any further.

jcheng5:
I've made some changes on my own branch: https://github.com/rstudio/leaflet/tree/draw-toolbar

This included squashing your original commits after re-roxygenizing. I think you're running a different version of roxygen2 which resulted in a lot of spurious Rd changes.

If you're OK with my changes I'll replace this PR with one based on that branch, just let me know what you think.

gherardovarando:
Yes of course, thanks a lot

byzheng:
Thanks for adding draw feature which I am looking for a few days. Do you have any plan to merge it into master?

PR #184, opened 2015-09-11
Integration with 5 Leaflet plugins

bhaskarvk:
Added support for the following Leaflet plugins
 - https://github.com/joergdietrich/Leaflet.Terminator
- https://github.com/ablakey/Leaflet.SimpleGraticule
- https://github.com/bbecquet/Leaflet.MagnifyingGlass
- https://github.com/Norkart/Leaflet-MiniMap
- https://github.com/lvoogdt/Leaflet.awesome-markers

All have examples under inst/example dir.
I bundled them all in one pull request coz they are small and standalone.


timelyportfolio:
:+1:

timelyportfolio:
@bhaskarvk in my RStudio Viewer on Windows 7, the magnifying glass is black and opaque.  It works fine however in Chrome and Firefox.

image

bhaskarvk:
Works for me. Rstudio 0.99.679, R 3.2.2
image

bhaskarvk:
@timelyportfolio Could you try the other two examples. One with the control and another with the layer, and toggling the control/layer to see if it makes any diff.

timelyportfolio:
Same result with all the examples.  I'll dig a little more.  I'm on RStudio 0.99.473 with R 3.2.2.  I'll update to RStudio nightly.

bhaskarvk:
@timelyportfolio Any updates with latest R-studio ? @jcheng5 Any interest in merging this PR ? I think these are some useful plugins to have.

timelyportfolio:
no, still black blob even with latest R-studio.  Just for reference, I think Joe is busy with Earl right now.

jcheng5:
I just got back from EARL, looking at this now and tomorrow. Thanks for your patience!

inst/examples/awesomeMarkers.R
```diff
@@ -0,0 +1,78 @@
...
+# Make a list of icons. We'll index into it based on name.
+popIcons <- awesomeIconList(
+  blue = makeAwesomeIcon(icon='users',prefix='fa', markerColor = 'blue'),
+  red = makeAwesomeIcon(icon='users',prefix='fa', markerColor = 'red')
+)
+
+leaflet(cities) %>% addTiles() %>%
+  addAwesomeMarkers(lng = ~Long, lat = ~Lat,
+             label = ~City,
+             labelOptions = rep(labelOptions(noHide = T),nrow(cities)),
```
jcheng5: This seems to be a less efficient version of `labelOptions = labelOptions(noHide = T)`, but perhaps you did this on purpose just to show you could have individual options if desired?

jcheng5:
This is impressive work, all the examples seem to work well.

Are there any areas you'd like feedback on, or parts of the code worth looking particularly closely at?

inst/examples/awesomeMarkers.R
bhaskarvk: Yes indeed, I figured the label options needed a bit more flexibility than popup options. So that you can decorate each label differently. Btw labelOption is part of the earlier Leaflet.label plugin that you've already merged into master. So you would do the same thing for adding labels to normal markers too.

bhaskarvk:
One thing I haven't really looked at in detail yet is the shiny part. I haven't yet picked up shiny yet, so I didn't check how these would work in shiny. But given that they either extend Marker or Layer, I suppose you can use the base class methods w/o having to add any explicit code.

I'm working on some other plugins as well, my humble contribution to what is already an awesome project.

bhaskarvk:
@jcheng5 I merged your PR. Funny I never ran in to that Rstudio issue.

jcheng5:
@bhaskarvk The RStudio issue would only repro on Windows and Linux. The Mac version of RStudio has used Safari as its rendering engine instead of QtWebkit since shortly after the release of Mavericks.

I'll make another PR to add Shiny features to the awesome marker integration, then I'd say we are probably good to go.

R/plugin-awesomeMarkers.R
```diff
@@ -0,0 +1,187 @@
+leafletAwesomeMarkersDependencies <- function() {
+  list(
+    htmltools::htmlDependency(
+      "leaflet-awesomemarkers",
+      "2.0.2",
+      system.file("htmlwidgets/plugins/Leaflet.awesome-markers", package = "leaflet"),
+      script = c('leaflet.awesome-markers.min.js','bootstrap.min.js'),
+      stylesheet = c('leaflet.awesome-markers.css','bootstrap.min.css',
```
jcheng5: The bootstrap and font-awesome dependencies need to be broken out in to separate `htmlDependency` objects. Otherwise this will break dependency resolution when leaflet is combined with Shiny.

This is what the htmlDependency objects look like in Shiny:
https://github.com/rstudio/shiny/blob/3d5c184acce71290533e8b3627638fca67fd9fd8/R/bootstrap.R#L39-L52
https://github.com/rstudio/shiny/blob/3d5c184acce71290533e8b3627638fca67fd9fd8/R/bootstrap.R#L1340-L1343

Let me know if you don't understand, I can make the change if you need me to.

jcheng5:
One last change I'd appreciate: For some of the functions, the parameter documentation is just a link to the JS docs. As tempting as this is, we try to avoid this, at least for the options we think people will actually use (it's OK to have a `...` argument that says "any other options, see docs for details"). This is not only to save the user from navigating through the docs, but especially so that autocomplete in RStudio can show a useful tip.

Thanks again for these fantastic changes, I'm really looking forward to the next release of this package!

R/plugin-awesomeMarkers.R
bhaskarvk: I can split them out, I even thought of doing that. But then it would mean I can't add them by default , right ? The dependency will have to be added by an extra function call , something like addBootStrap() / addFontAwesome() , right ?

bhaskarvk:
@jcheng5 I'll make all these changes over the weekend. Thanks for all the pointers.

R/plugin-awesomeMarkers.R
jcheng5: You can decide how you want to do it. If you want to add it by default, you can; just have this list contain all three htmlDependency objects, instead of just the one.

But you could also detect if "glyphicon" or "fa" is being used, and add just font-awesome or bootstrap instead of both. This wouldn't go in a new place in the code, it would go at the top of `addAwesomeMarkers` where you're already adding stuff to `map$dependencies`.

bhaskarvk:
@jcheng5 Sorry for the delay but now everything is as per your request. Please review and merge.

jcheng5:
Thanks so much, I'll take a look as soon as I can.

jcheng5:
Other than the fixes I posted in https://github.com/bhaskarvk/leaflet/pull/6, this looks good to me!

PR #204, opened 2015-11-03
Leaflet 0.7.7

bhaskarvk:
Upgraded leaflet 0.7.7.


inst/htmlwidgets/lib/leaflet/leaflet.css
```diff
@@ -70,7 +71,7 @@
 	height: 1px;
 	}
 .lvml {
-	behavior: url/**/(#default#VML);
+	behavior: url(#default#VML);
```
jcheng5: This change needs to be undone; the `/**/` is necessary to prevent Pandoc from choking when Leaflet is used in an R Markdown document.

jcheng5:
Thanks for this!

bhaskarvk:
No worries, it's a very small commit anyways.

bhaskarvk:
@jcheng5 This may not be fully ready. See https://github.com/rstudio/leaflet/issues/206.

Issue #209, opened 2015-11-24
Feature Request: Marker Popup Open/Close


ReportMort:
It would be nice to be able to programmatically open or close popups bound to existing markers. My current workaround is to use `addPopups()` at the same lat/lng of an existing marker, which doesn't seem ideal considering that leaflet markers have an methods to open and close popups (http://leafletjs.com/reference.html#marker-openpopup).

PR #210, opened 2015-11-24
Add Marker Popup Methods

ReportMort:
Expose marker popup methods so that bound marker popups can be programmatically controlled. This is in reference to the feature request mentioned in #209.

I have setup a variety of examples that demo the proposed methods:
- A set of demos/tests:
  - https://gist.github.com/ReportMort/2af90ea619ec564d7174
- A Shiny gist
  - https://gist.github.com/ReportMort/6e457e92bb65f437dff5
- A deployed sample Shiny application:
  - https://reportmort.shinyapps.io/leaflet-marker-methods-example

The leaflet methods being replicated are located starting here: http://leafletjs.com/reference.html#marker-unbindpopup


bhaskarvk:

jcheng5:
I feel like doing this properly would mean not just supporting it for marker popups, but all layer-attached popups. And I also think if you're involving Shiny, you can already do this after a fashion by not having marker popups at all but adding the popups when the marker is clicked; then the popup has its own layerId and can be controlled using `removePopup`.

ReportMort:
@jcheng5 I would be in favor of adding these methods for all layer-attached popups. I made the PR since I wasn't really satisfied with using add/remove popup as a workaround. `addPopups` and `removePopup` requires fiddling with the lat/long to get their placement over the marker, the popups don't autoclose when other marker popups are clicked, and must be micro-managed outside of the map data.

bhaskarvk:
FWIW I really like this PR, and making it work for all shapes is fairly easy. I might give it a shot next week.

ReportMort:
Thanks, let me know if you need anything or I need to reopen. I closed because I got the impression that it wasn't something people wanted in the package.

PR #217, opened 2015-12-06
Add support for SpatialMultiPoints*

kent37:
I added support for SpatialMultiPoints and SpatialMultiPointsDataFrame (from package sp) to addPolylines() and addPolygons() by specializing polygonData for these two classes. No tests but I am using it...

### DIFF
--- a/R/leaflet.R
+++ b/R/leaflet.R
@@ -14,8 +14,9 @@
 #'   frames, and spatial objects from the \pkg{sp} package of classes
 #'   \code{SpatialPoints}, \code{SpatialPointsDataFrame}, \code{Polygon},
 #'   \code{Polygons}, \code{SpatialPolygons}, \code{SpatialPolygonsDataFrame},
-#'   \code{Line}, \code{Lines}, \code{SpatialLines}, and
-#'   \code{SpatialLinesDataFrame})
+#'   \code{Line}, \code{Lines}, \code{SpatialLines},
+#'   \code{SpatialLinesDataFrame}, \code{SpatialMultiPoints}
+#'   and \code{SpatialMultiPointsDataFrame})
 #' @param width the width of the map
 #' @param height the height of the map
 #' @param padding the padding of the map
--- a/R/normalize.R
+++ b/R/normalize.R
@@ -195,6 +195,16 @@ polygonData.SpatialPolygons = function(obj) {
 polygonData.SpatialPolygonsDataFrame = function(obj) {
   polygonData(sp::polygons(obj))
 }
+polygonData.SpatialMultiPoints = function(obj)
+{
+  lapply(obj@coords, function(points) list(list(lng=points[,'lon'], lat=points[,'lat']))) %>%
+    structure(bbox = obj@bbox)
+}
+polygonData.SpatialMultiPointsDataFrame = function(obj)
+{
+  polygonData(sp::SpatialMultiPoints(obj@coords))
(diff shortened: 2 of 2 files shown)


jcheng5:
Thanks for the contribution!

`SpatialMultiPoints` is new to me. It looks like it represents distinct groups of points. Is it really appropriate to have this work with `addPolylines()` and `addPolygons()`? Doesn't `addMarkers()` and `addCircleMarkers()` make more sense?

jcheng5:
For my own reference, sample code from Edzer:

```r
cl1 = cbind(rnorm(3, 10), rnorm(3, 10))
cl2 = cbind(rnorm(5, 10), rnorm(5,  0))
cl3 = cbind(rnorm(7,  0), rnorm(7, 10))

library(sp)
mp = SpatialMultiPoints(list(cl1, cl2, cl3))
plot(mp, col = 2, cex = 1, pch = 1:3)
mp
mp[1:2]

print(mp, asWKT=TRUE, digits=3)

mpdf = SpatialMultiPointsDataFrame(list(cl1, cl2, cl3),
data.frame(a = 1:3))
mpdf

plot(mpdf, col = mpdf$a, cex = 1:3)
as(mpdf, "data.frame")
mpdf[1:2,]
```

kent37:
That's a good question. I am using `SpatialMultiPointsDataFrame` to hold three-dimensional track data - lat, lon and altitude. (I can't use `SpatialLines`, it is 2-d only.) SMPDF was suggested by rspatial folks so I am not abusing it too much :-). My data is airplane tracks. Among other things I want to show them as lines on a map so `addPolylines()` works for me.

I can imagine an application that would want to use `addCircles()` or `addCircleMarkers()` with `SpatialMultiPoints` but that is not my need.

bhaskarvk:

jcheng5:
@kent37 does SpatialMultiPoints really make more sense than SpatialLinesDataFrame with an altitude attribute? Or just a regular data frame?

kent37:
@jcheng5 I have moved to just using a data frame. I'm OK with not fixing this one.

mdsumner:
I am very sympathetic to this kind of abuse of the spatial definitions, you can store more attributes than 2 (x-y) on the coordinates for Points and MultiPoints in sp, but not so lines/polygons. This is really the reason why there are dozens of "tracking" packages on CRAN, you want at least X-Y-Z, then Time, and possibly Temperature, etc.  These values belong on the vertices, not on the features (i.e. the track line).  The new sf framework allows for Z and/or M as well as X-Y, but removes this more general possibility for Points. I wrote https://github.com/r-gris/rangl to allow for more general models, but it's not obvious where this should go, it's only really able to store and plot for now.  R's leaflet will want a more general model than Spatial provides, but there's no clear general framework to build on. The mapview package faces similar discussions. 

FWIW, these kinds of problems come up in QGIS, and they have "geometry generators" now that allow manipulating the underlying primitives (line segments, vertices) of a line feature. But still it's not a general framework, with features hierarchically composed of primitives and so on. R could use the ggplot2 approach with aesthetics and grouping, but it needs to be defined more generally.

bhaskarvk:
I am closing this. With `sp` on its way out we should concentrate our efforts on supporting `sf` correctly.

kent37:
Agree on concentration on sf. I am moving my own code to using sf which has good support for linestrings with xyzm coordinates.

Issue #224, opened 2015-12-31
Strange behavior from `colorNumeric` & `colorFactor` clipping raster values (setting as NA?).


SimonGoring:
I am using leaflet to represent some spatial data (historical tree species distributions - see [preprint](http://biorxiv.org/content/early/2015/10/24/026575) here for more information) as part of a bigger shiny app.  The branch I'm working on "[leaflet](https://github.com/PalEON-Project/Shiny/tree/leaflet)" contains the code.

There's lots to the code, but of note is that both `colorNumeric` and `colorFactor` (around line 154 in `server.R`) are exhibiting some strange behavior.  Both appear to be returning:

```
Warning in colors(.) :
  Some values were outside the color scale and will be treated as NA
```

even when I've tried to set the ranges to be greater than (or equal to) the range of the data.  I don't think it's a problem on my end, since the clipping is happening in very particular places.  In a strange geometric pattern for `colorNumeric` (see figure below) and at the boundaries between discrete levels for `colorFactor`.

Any help would be appreciated.
![leaflet_colornumeric](https://cloud.githubusercontent.com/assets/1619126/12060244/34e41b7e-af20-11e5-8d1b-b52640b36aab.png)
The `colorNumeric` clipping.

![leaflet_colorfactor](https://cloud.githubusercontent.com/assets/1619126/12060289/4d6e47ea-af21-11e5-91f5-b36c9029d64d.png)
The `colorFactor` clipping.

jcheng5:
Is this happening with `addRasterImage`? That function reprojects the raster to match with the projection that Leaflet uses; see `leaflet::projectRasterForLeaflet`. IIRC I've found that reprojecting seems to cause the interpolated values to sometimes exceed the original max/min, I'm not sure why. Can you try running your raster through `leaflet::projectRasterForLeaflet` and see if that might be the case here?

SimonGoring:
Yes, I've managed to fix it.  

It's happening (apparently) because even if the raster is in the native leaflet projection (which I assume is `EPSG:3857`) `addRaster` reprojects, so I've only had to add `project=FALSE`.  However, this behavior isn't described in the help, so it's not clear to me what's actually happening.  

For `colorFactor` it's reprojecting and resampling using some form of interpolation, so the edges of each class are getting lost (since they don't fit into the palette).  For `colorNumeric` it's not clear to me why some values are apparently higher than they were originally.

[...]

Issue #219, opened 2015-12-11
`projectRasterForLeaflet` should support `method="ngb"`


adrfantini:
The `raster` package supports nearest neighbour projection in `projectRaster`, by issuing `method="ngb"`. This should be possible in `projectRasterForLeaflet` too - it's useful for a number of different datasets.

tim-salabim:
I totally agree that there needs to be an option for re-projecting factorial rasters using nearest neighbor. The linear interpolation of the bilinear re-projection (the default in `raster::projectRaster` produces floating point values which, in addition to being false values for factorial raster layers, also mess up the legend produced with `leaflet::colorFactor`.

In mapview we actually do the raster projection outside the `leaflet::addRasterImage` call like this (lines 178 - 187 in https://github.com/environmentalinformatics-marburg/mapview/blob/master/R/leafletControls.R):

```r
is.fact <- raster::is.factor(x)[1]
if (is.fact) {
    x <- raster::projectRaster(
      x, raster::projectExtent(x, crs = sp::CRS(wmcrs)),
      method = "ngb")
    x <- raster::as.factor(x)
  } else {
    x <- raster::projectRaster(
      x, raster::projectExtent(x, crs = sp::CRS(wmcrs)),
      method = "bilinear")
  }
```

The user still needs to make sure that their raster is of class factor (as having integer values does not necessary mean factorial, hence the explicit call to `raster::as.factor` after the re-projection), but that can be easily done with `raster::as.factor`

PR #228, opened 2016-01-15
Two key changes to address issues #224 and #219 (reprojecting raster for `addRaster`)

SimonGoring:
1.  Added `method` to the options so that users can define eother `'ngb'` or `'bilinear'` options for `raster::projectExtent` to improve plotting of categorical data. (and added to `roxygen2` documentation block)

2.  Do not re-project if CRS for the raster is equivalent to the current CRS.  I'm not sure if this is that common an issue, but just to avoid any potential (accidental) reprojection (see issue #224).


jcheng5:
Did you change your mind?

SimonGoring:
It fails the check.  I'm going to post a new pull request.

PR #229, opened 2016-01-18
Added the "method" variable to `addRasterImage`

SimonGoring:
This is in reference to issues #219 and #224.  I tried to figure out a way to make the direct comparison between projections, but the `epgs3857` used here is not (strictly speaking) the one unpacked by `rgdal` when you use `+init=epsg:3857`.

### DIFF
--- a/R/layers.R
+++ b/R/layers.R
@@ -223,9 +223,7 @@ addRasterImage = function(
 #' @rdname addRasterImage
 #' @export
 projectRasterForLeaflet <- function(x, method) {
-  if(!raster::crs(x) == sp::CRS(epsg3857)){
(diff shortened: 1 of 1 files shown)


bhaskarvk:

jcheng5:
I guess so. In the browser the image will be drawn with bilinear or nearest neighbor depending on zoom level. But I think that's probably fine.
https://github.com/rstudio/leaflet/blob/e38c94837cd1a8e336a5a3aa75daa38a314d2eff/inst/htmlwidgets/leaflet.js#L1836-L1847

We will need to get a CLA from @SimonGoring.

bhaskarvk:
This is such a small PR do we really need a CLA signed for this ? Other than that looks like something we should include right away.

bhaskarvk:
Duplicate of #462

PR #263, opened 2016-05-12
choropleth working example

bbest:
Hi @jcheng5,

Here is a choropleth example using Shiny which works with the latest rstudio/leaflet package (vs the old legacy [choropleth](https://github.com/rstudio/leaflet/tree/master/inst/legacy/examples/choropleth)) and is meant to replicate the functionality of the [Interactive Choropleth Map at LeafletJS.com](http://leafletjs.com/examples/choropleth.html).

Here is a screenshot hovering the cursor over California:

leaflet_bbest_choropleth

This improves on the legacy version by now using the following features:
- `AddPolygons(layerId)` to gain access to `input$map1_shape_mouseover$id`, per [Leaflet for R - Shiny - Inputs/Events](http://rstudio.github.io/leaflet/shiny.html#inputsevents)
- `clearGroup("highlight")` for removing any previously highlighted shape. Strangely, `removeShape()` did not work.
- added some logic to clear highlighted shape after mouseout
- some simplification of code to generate break labels for legend and use RColorBrewer color palette vs manual specification of hexadecimal codes

To try this yourself:

``` r
devtools::install_github('bbest/leaflet@feature/choropleth')
library(leaflet)
print(source(system.file('examples/shiny-choropleth.R', package='leaflet'))$value)
```


bhaskarvk:
I am going to close this. We don't need Shiny for the hover effects any more. `add<Shape>` methods all support highlightOptions for this.

PR #269, opened 2016-05-27
Add `bringTo` bindings

nikolai-b:
I know these could be done easily through the `invokeMethod` but this is nicer.

I've not runroxygen2 but can if you'd like.


bhaskarvk:
Closing as no demand from anyone else, also the PR is incomplete. The JS side is missing.

PR #294, opened 2016-09-18
Feature/plugin proj4 leaflet

bhaskarvk:
@jcheng5 Not for merging, only for review.

Code sample http://rpubs.com/bhaskarvk/proj4leaflet.

cc @timelyportfolio / @tim-salabim.


jcheng5:
Try running `git-apply` this patch: https://gist.github.com/jcheng5/4167fadc93be166cae94de13dba3b029/raw/6ba0dd73c7deecc61b6e8f14a443bb69082a4955/onrender.diff

This is one approach we can take to preserving backward compatibility; it doesn't involve changing htmlwidgets, but it does make an assumption about how render hooks are stored on the htmlwidget object.

R/leaflet.R
```diff
@@ -19,15 +19,24 @@
...
 #' @param height the height of the map
 #' @param padding the padding of the map
+#' @param mapOptions the map options
 #' @return A HTML widget object, on which we can add graphics layers using
 #'   \code{\%>\%} (see examples).
 #' @example inst/examples/leaflet.R
 #' @export
-leaflet = function(data = NULL, width = NULL, height = NULL, padding = 0) {
+leaflet = function(data = NULL, width = NULL, height = NULL,
+                   padding = 0, mapOptions = list()) {
```
jcheng5: I'd like you to rename the `mapOptions` argument as `options` (and make its default `mapOptions()`), and add a `mapOptions` function that takes crs/minzoom/maxzoom/maxbounds as arguments (center and zoom are covered by setView--or if you prefer we could have center/zoom in both places). This style is what we've used for the other functions and lets us provide autocomplete help.

R/leaflet.R
```diff
@@ -36,14 +45,37 @@ leaflet = function(data = NULL, width = NULL, height = NULL, padding = 0) {
...
 
+hookWrapperTemplate <- "function(el, x, data) {
+  return (%s).call(this.getMap(), el, x, data);
+}"
+
 getMapData = function(map) {
   attr(map$x, "leafletData", exact = TRUE)
 }
 
+getMapOptions = function(map) {
```
jcheng5: Is there a reason you added this? It's not referenced anywhere else that I can tell.

R/leaflet.R
```diff
@@ -70,3 +102,71 @@ mapOptions <- function(map, zoomToLimits = c("always", "first", "never")) {
 
   map
 }
+
+# CRS classes supported
+crsClasses <- list( 'L.CRS.EPSG3857', 'L.CRS.EPSG4326', 'L.CRS.EPSG3395',
+                    'L.CRS.Simple', 'L.Proj.CRS', 'L.Proj.CRS.TMS' )
```
jcheng5: Just checking--we'll be able to support (at least) these CRS classes even after we move to Leaflet 1.0?

javascript/src/index.js
```diff
@@ -72,160 +73,356 @@ function preventUnintendedZoomOnScroll(map) {
...
-    // there.
-    if (el.offsetWidth === 0 || el.offsetHeight === 0) {
-      map.leafletr.pendingRenderData = data;
-      return;
-    }
-    map.leafletr.pendingRenderData = null;
 
-    // Merge data options into defaults
-    let options = $.extend({ zoomToLimits: "always" }, data.options);
+        map = L.map(el, data.options);
```
jcheng5: Before doing this, do you have to check if there's an existing map and destroy it if so?

javascript/src/index.js
```diff
@@ -72,160 +73,356 @@ function preventUnintendedZoomOnScroll(map) {
...
+          if (map.leafletr.pendingRenderData) {
+            this.doRenderValue(map.leafletr.pendingRenderData, map);
+          }
+        }
+      }
+    };
   }
 });
 
+//###HTMLWidgets.widget({
```
jcheng5: Let's remove all this :)

R/leaflet.R
bhaskarvk: Agreed on renaming mapOptions argument to options.

 But as for introducing a mapOptions() function, there's already a mapOptions function, which behaves differently than other options() function.  tileOptions(), markerOptions() etc. all give back a list . But mapOptions as it's implemented now returns a map. Here's the current implementation

```r
function (map, zoomToLimits = c("always", "first", "never"))
{
    if (is.null(map$x$options))
        map$x$options <- list()
    zoomToLimits <- match.arg(zoomToLimits)
    map$x$options$zoomToLimits <- zoomToLimits
    map
}
```

If I make it similar to other Options() functions it breaks compatibility.

R/leaflet.R
bhaskarvk: If it's OK with you I would rather put it in, it may come handy in certain use cases, none of which are coded in as of now. Besides it makes the code around the map object symmetric.

R/leaflet.R
bhaskarvk: Yes, but on a side note this is not an exported list, it is used internally for checking the crsClass argument to mapOptions. So technically it can change w/o breaking API.
But just to be sure, I checked 1.0 code and these projs. are indeed supported there.

javascript/src/index.js
bhaskarvk: will do. I need to read up on htmlwidgets code to truly understand what happens when, and scenarios where there will already be an instance of map.

javascript/src/index.js
bhaskarvk: Yup, I left it for the code review part, coz git diff was wanky for this portion, so easier to check new code against commented code in same file. :)

bhaskarvk:
Closing this one, preparing a new one as per discussion.

bhaskarvk:
Now part of #298.

PR #298, opened 2016-09-22
2 new plugins Proj4Leaflet, EasyButton and Leaflet upgrade from 0.7.3 to 0.7.7

bhaskarvk:
@jcheng5 All ready for merging based on our discussion. Leaflet also upgraded to 0.7.7


R/plugin-easybutton.R
```diff
@@ -0,0 +1,186 @@
...
+#' Creates an easy button.
+#' @seealso \url{https://github.com/CliffCloud/Leaflet.EasyButton}
+#' @param icon the button icon
+#' @param title text to show on hover
+#' @param onClick the action to take
+#' @param position topleft|topright|bottomleft|bottomright
+#' @param id id for the button
+#' @param states the states
+#' @export
+easyButton <- function(
```
jcheng5: It looks like you're using parameter defaults as documentation. Can you not do that? The default values should be an actual reasonable default that the user would want, or nothing. In this case definitely the default `title` and `onClick`, at least, should be nothing--we want to force people to provide those.

(Only exception being a vector of possible choices, but even then, only if the first element is actually a reasonable default value.)

R/plugin-easybutton.R
```diff
@@ -0,0 +1,186 @@
...
+#'   addTiles() %>%
+#'   addEasyButton(easyButton(
+#'      icon = htmltools::span(class='star','&starf;'),
+#'      onClick = JS("function(btn, map){ map.setZoom(1);}")))
+#'
+#' @describeIn easyButton add an EasyButton to the map
+#' @export
+addEasyButton <- function(
+  map,
+  button = easyButton()
```
jcheng5: I think this `button` param also shouldn't have a default value.

jcheng5:
Other than the remaining nits, LGTM!

Assuming you're OK with my suggested changes, you can fix those then merge.

R/plugin-easybutton.R
bhaskarvk: Oh Shoot that was just some dummy code I had setup for testing purposes. I'll take it out.

PR #323, opened 2016-11-03
Some more small features

bhaskarvk:
@jcheng5 . One last PR before 1.1.

- Exported some useful dependency functions.

- Added a way to specify label, popup and highlightOptions to GeoJSON and TopoJSON.
I purposefully used the argument names 'labelProperty' and 'popupProperty' instead of the 'label' and 'popup'  argument names in other add* methods. This was to distinguish from how these params are interpreted in the addGeoJSON/addTopoJSON functions as opposed to the other add*  fuctions. But the API is fairly R idiomatic (I think).

You can see the GeoJSON handling with 3 approaches at
http://rpubs.com/bhaskarvk/geojson-handling

And TopoJSON's with labels/popups and highlights at
http://rpubs.com/bhaskarvk/topoJSON


javascript/src/methods.js
```diff
@@ -543,6 +543,43 @@ methods.addGeoJSON = function(data, layerId, group, style) {
...
     data = JSON.parse(data);
   }
+  let popupProperty = style.popupProperty;
+  let labelProperty = style.labelProperty;
+  delete style.popupProperty;
+  delete style.labelProperty;
+  let popupOptions = style.popupOptions;
+  let labelOptions = style.labelOptions;
+  delete style.popupOptions;
+  delete style.labelOptions;
```
jcheng5: Better to pass these values as a separate object rather than piggyback them onto style and then pull them off. Same with highlightStyle below.

javascript/src/methods.js
```diff
@@ -543,6 +543,43 @@ methods.addGeoJSON = function(data, layerId, group, style) {
...
+
+  let defaultStyle = {};
+  let highlightStyle = style.highlightOptions;
+  delete style.highlightOptions;
+  if(!$.isEmptyObject(highlightStyle)) {
```
jcheng5: It looks like the highlightStyle iteration, highlightFeature, and resetFeature are all copy/pasted from addLayers. Please refactor.

javascript/src/methods.js
```diff
@@ -600,8 +722,55 @@ methods.addTopoJSON = function(data, layerId, group, style) {
         featureId: feature.id,
         properties: feature.properties
       };
-      let popup = feature.properties.popup;
-      if (typeof popup !== "undefined" && popup !== null) layer.bindPopup(popup);
+      //let popup = feature.properties.popup;
```
jcheng5: Ditto to the copy/pastes in addTopoJSON.

R/layers.R
```diff
@@ -1086,6 +1086,8 @@ clearShapes = function(map) {
 
 #' @param geojson a GeoJSON list, or character vector of length 1
 #' @describeIn map-layers Add GeoJSON layers to the map
+#' @param popupProperty The property to use for popup content
+#' @param labelProperty The property to use for the label.
```
jcheng5: Can you document that these can either be the names of the properties, or JavaScript functions? With an example?

bhaskarvk:
I'll refactor accordingly

bhaskarvk:
Closing as per discussion.

PR #329, opened 2016-11-19
Instanciate map via JS factory method.

bhaskarvk:
Also changed the group element from layerGroup to featureGroup so that it's compatible with Draw.

I also added lodash as a core JS dependency. Al though there's no existing code using loadsh is leaflet, I plan to do a lot of functional FP in down stream packages, so makes sense to have it at leaflet level.


bhaskarvk:
@jcheng5 Please don't merge until I give a green light.

bhaskarvk:
I took out lodash JS as it was slowing down the widget loading time. Now only with changes that allow custom map initialization. This is good to go.

Issue #213, opened 2015-12-02
addPolylines does not support lines broken in several pieces


jiho:
This example has two lines identified separately in a `SpatialLines` object and it works
```r
x1 <- SpatialLines(
  LinesList = list(
    Lines( slinelist=list( Line(cbind(c(5,6,7),c(43,44,45))) ), ID=1),
    Lines( slinelist=list( Line(cbind(c(8,9,9),c(43,44,45))) ), ID=2)
  )
)
leaflet() %>% addTiles() %>% addPolylines(data=x1)
```

Now with two pieces of a line with the same ID, `addPolylines` only shows the first piece.
```r
x2 <- SpatialLines(
  LinesList = list(
    Lines( slinelist=list(
        Line(cbind(c(5,6,7),c(43,44,45))),
        Line(cbind(c(8,9,9),c(43,44,45)))
      ), ID=1
    )
  )
)
leaflet() %>% addTiles() %>% addPolylines(data=x2)
```

Getting pieces of lines is usual when computing them (using `rasterToContour` for example).

There is probably a way to convert from
```r
leaflet:::polygonData.SpatialLinesDataFrame(x2)
```
to
```r
leaflet:::polygonData.SpatialLinesDataFrame(x1)
```
But I am guessing there is probably something more clever to be done!

NB: given that the code is much of the same, I suspect this bug is also true for polygons which are composed of several sub-polygons.

coolbutuseless:
+1 for this problem.   I went to the trouble of creating a minimal example, and so will leave it here for posterity

```
library(dplyr)
library(leaflet)
library(sp)

packageVersion('sp')
packageVersion('leaflet')

#----------------------------------------------------------
# Vertical and horizontal line to make
# left side and top of triangle
# i.e.    _____
#         |
#         |
#
#----------------------------------------------------------
coords1a <- matrix(c(
  0, 0,
  0, 1,
  1, 1
), nrow = 3, byrow = TRUE)


#----------------------------------------------------------
# Diagonal line to complete triangle i.e.
#
#              /
#             /
#            /
#
#----------------------------------------------------------
coords1b <- matrix(c(
  0, 0,
  1, 1
), nrow = 2, byrow = TRUE)


#----------------------------------------------------------
# Create an sp::SpatialLines object
#----------------------------------------------------------
line1a   <- Line(coords1a)
line1b   <- Line(coords1b)
S1       <- Lines(list(line1a, line1b), ID="tri")
[...]

PR #332, opened 2016-11-22
fixed rstudio/leaflet#213

bhaskarvk:
This was a small fix and the workaround documented in #213 is more complicated than the fix.

### DIFF
--- a/javascript/src/methods.js
+++ b/javascript/src/methods.js
@@ -131,7 +131,6 @@ function addMarkers(map, df, group, clusterOptions, clusterId, markerFunc) {
     let clusterGroup = this.layerManager.getLayer("cluster", clusterId),
       cluster = clusterOptions !== null;
     if (cluster && !clusterGroup) {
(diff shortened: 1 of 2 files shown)


javascript/src/methods.js
```diff
@@ -442,9 +439,15 @@ methods.addPolylines = function(polygons, layerId, group, options, popup, popupO
       .cbind(options);
 
     addLayers(this, "shape", df, function(df, i) {
-      let shape = df.get(i, "shapes")[0];
-      shape = HTMLWidgets.dataframeToD3(shape);
-      return L.polyline(shape, df.get(i));
+      let shapes = df.get(i, "shapes");
```
jcheng5: It feels weird to mutate the shape array you get back from df. Instead of doing this, can you use `shapes.map(HTMLWidgets.dataframeToD3)` or similar?

bhaskarvk: Haha I did use `map` first and then noticed the for loop in `addPolygons`, so I changed it to for loop to keep it consistent.

PR #343, opened 2016-12-20
relative urls for openstreetmap tiles

mhines-usgs:
suggested change for either http or https for openstreetmap requests using relative urls

### DIFF
--- a/R/layers.R
+++ b/R/layers.R
@@ -112,7 +112,7 @@ hideGroup <- function(map, group) {
 #' @export
 addTiles <- function(
   map,
-  urlTemplate = 'http://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png',
+  urlTemplate = '//{s}.tile.openstreetmap.org/{z}/{x}/{y}.png',
   attribution = NULL,
   layerId = NULL,
   group = NULL,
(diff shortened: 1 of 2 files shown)


bhaskarvk:
+1 for this PR.

mhines-usgs:
Any ETA on merging this? or other suggestions to hurry it along? Thanks :D

isuftin:
👍

jcheng5:
Thanks for the PR! And sorry for the delay.

bhaskarvk:
Actually there is a problem w/ using protocol relative url. If you do

```r
l <- lealfet() %>% addTiles()
htmlwidgets::savewidget(l, 'l.html')
```

And open the file in a browser over the `file://` protocol, the `//` of the tile URL template becomes `file://` and thus no tiles are loaded.

mhines-usgs:
That's a bummer! I didn't realize that was a use case (local file://) 

I can't think of a workaround for that need. Do you have suggestions for avoiding mixed content warnings where this is hardcoded to http and being used on an https website? 

There are other places in the package that have relative URLs, do those not cause issues with file:// access?

This package is pulled in at build currently, my other idea is forking it and pointing at that fork instead of the master repo, if need be.

jcheng5:
Crap, I missed this issue. The fix is to simply use https I think.

PR #348, opened 2017-01-06
Export all normalize methods

hadley:
I don't understand when this is necessary, but I definitely odd bugs when not exported.


bhaskarvk:
+1 for this PR, It will help me in leaflet.extras too.

PR #354, opened 2017-01-09
support MULTIPOLYGON by ruthless flattening

mdsumner:
This pull request adds support for MULTIPOLYGON, by downgrading the extra structure in the `sf` object to one analogous to that used by `sp` (and shapefiles).  I.e. rings that are holes in other rings are stored in one flat collection, ambiguously about which hole belongs where. This works for vis. though because the polygon rendering just uses the evenodd rule. 

It's a problem for round-tripping, but then on the `sf` side it's not going to be hard to calculate which ring belongs in which island. I thought it should be discussed at least, but very happy to see this supported now, thanks!  

Basic example, from the sf1 vignette: 

```R
## multipolygon
library(sf)
p1 <- rbind(c(0,0), c(1,0), c(3,2), c(2,4), c(1,4), c(0,0))
p2 <- rbind(c(1,1), c(1,2), c(2,2), c(1,1))
pol <-st_polygon(list(p1,p2))
p3 <- rbind(c(3,0), c(4,0), c(4,1), c(3,1), c(3,0))
p4 <- rbind(c(3.3,0.3), c(3.8,0.3), c(3.8,0.8), c(3.3,0.8), c(3.3,0.3))[5:1,]
p5 <- rbind(c(3,3), c(4,2), c(4,3), c(3,3))
(mpol <- st_multipolygon(list(list(p1,p2), list(p3,p4), list(p5))))

mp <- st_sfc(mpol, crs = "+proj=merc +datum=WGS84")
library(leaflet)
leaflet() %>% leaflet::addPolygons(data = mp)

```

Here are two real-world data sets with different map projections, and interestingly different situations of multipolygons, multiple features, and holes. 

There's code here to download two GeoPackage files (850Kb and 1.4Mb), transform them to angular coordinates on the WGS84 datum, and plot.  I include a subsetting, because Lake Superior proper has holes in it, that are filled by other features.  

```R
library(leaflet)
library(sf)
## a couple of realer examples 850KB and 1.4Mb 
uf <- c("https://github.com/r-gris/polyggon/raw/master/inst/extdata/water_lake_superior_basin.gpkg", 
        "https://github.com/r-gris/polyggon/raw/master/inst/extdata/inlandwaters.gpkg")
td <- tempdir()
for (i in seq_along(uf)) download.file(uf[i], file.path(td, basename(uf[i])), mode = "wb")

## read the layers from the GeoPackage (it's a SQLite db, understood by GDAL)
lake_superior <- st_read(file.path(td, basename(uf[1])), layer = "lake_superior_basin")
lakey <- st_read(file.path(td, basename(uf[2])), layer = "inlandwaters")

## reproject to longlat :|
lake_superior_ll <- st_transform(lake_superior, "+proj=longlat +datum=WGS84")
lakey_ll <- st_transform(lakey, "+proj=longlat +datum=WGS84")

## Lake Superior
## 
## the holes have other features in them :)
leaflet() %>%  addPolygons(data = lake_superior_ll)
## this is the lake, which isn't a hole, but has holes in it, and there are islands in the holes :0
idx <- which.max(st_area(lake_superior))
leaflet() %>%  addPolygons(data = lake_superior_ll[idx, ])

## Inland waters, Australia
##
## these multipolygons with holes also work fine: 
leaflet() %>%  addTiles() %>% addPolygons(data = lakey_ll)
```

### DIFF
--- a/R/normalize-sf.R
+++ b/R/normalize-sf.R
@@ -23,6 +23,10 @@ pointData.sfc_POINT <- function(obj) {
   )
 }
 
+#' @export
+pointData.sfc_MULTIPOINT <- pointData.sfc_POINT
+
+
 # polygonData -------------------------------------------------------------
 
 #' @export
--- a/R/normalize-sf.R
+++ b/R/normalize-sf.R
@@ -44,13 +44,8 @@ polygonData.sfc <- function(obj) {
 #' @export
 polygonData.MULTIPOLYGON <- function(obj) {
   n <- vapply(obj, length, integer(1))
-  if (any(n > 1L)) {
-    warning(
-      "leaflet currently does not support MULTIPOLYGONS. Taking first",
-      call. = FALSE)
-  }
 
-  lapply(obj, function(x) sf_coords(x[[1]]))
+  unlist(lapply(obj, function(x) lapply(x, sf_coords)), recursive = FALSE)
 }
 #' @export
 polygonData.MULTILINESTRING <- polygonData.MULTIPOLYGON


mdsumner:
Added support for MULTIPOINT in response to 

https://github.com/rstudio/leaflet/issues/352

Following the example above: 

```R

library(dplyr)
leaflet() %>%  addTiles() %>% addCircles(data = st_cast(lakey_ll %>% filter(Province == "Tasmania"), "MULTIPOINT"))
```

Be careful with addMarkers and MULTIPOINT, it's really easy to overload the request casting from complex layers  :|

hadley:
I don't think it's worth doing a half-hearted fix when doing it properly is not much extra work.

mdsumner:
A proper fix (an extra level of hierarchy in the lists, or a different design altogether) has implications across the board for all uses of addPolygons, and other providers. I asked about these implications and got no response.  It's not half-hearted btw, this is exactly what I would do given the design (lists of coordinates) being used. What am I missing?

hadley:
We need `addMultiPolyons()` and associated js infrastructure support. It's not much work and it would better to handle multi-polygons 100% correctly.

mdsumner:
Ok, right. I'll have a look (I just learnt an enormous amount about how leaflet works).

bhaskarvk:
I was just going to suggest what @hadley said. Leaflet supports MultiPolygons, we just need to add support for them.

mdsumner:
Hi @bhaskarvk @hadley 
 I am in furious agreement, please consider this a hearty learning experience on my part! Sorry for the diversion, if nothing else I hope those example files are useful for testing. I'm very keen to learn to do this but it will take me some time, and is probably trivial for someone else.

jcheng5:
Sorry for being MIA on all leaflet and all spatial conversations lately--I've been entirely consumed with getting ready for rstudio::conf. A week from Monday is the soonest I can start paying attention.

Sounds like we can close this PR though?

mdsumner:
Yes indeed.

bhaskarvk:
Going thru some old closed PRs to ensure we've properly addressed these. @mdsumner / @tim-salabim, has this been addressed, or are we still missing proper multipolygon, multipoint, multiline support?
If so I would appreciate a quick test case for each of these cases, so I can begin to add the support and quickly test it out.

tim-salabim:
`MULTIPOINTS` are currently not supported, the other `MULTI*` features are:

```r
library(mapview)
library(sf)

# MULTIPOINT doesn't work
mpt = st_cast(breweries[1:5, ], "MULTIPOINT")
class(st_geometry(mpt))
leaflet() %>% addTiles() %>% addCircleMarkers(data = mpt)

# MULIPOLYGON works
mpl = franconia[1:5, ]
class(st_geometry(mpl))
leaflet() %>% addTiles() %>% addPolygons(data = mpl)
mapview(franconia[1:5, ]) # hover over polygons 2 and 5 to see highlighting of multi

# MULTILINE works
mln = st_transform(trails[1:5, ], crs = 4326)
class(st_geometry(mln))
leaflet() %>% addTiles() %>% addPolylines(data = mln)
mapview(mln) # hover e.g. line 4
```

bhaskarvk:
OK cool, MultiPoint can't really be supported as it's a limitation of leaflet. We can hack support for it, but I would rather not, if the demand for it is not high.

mdsumner:
I put the previous examples together there, fwiw: http://rpubs.com/cyclemumner/292067  

One minor point, if this error could say as much as you did here that will help (I'd forgotten about multipoint being unsupport-able). 

Modify this to something like this so the xp is direct. 

"Please note that this is a fundamental limitation of the leaflet library ... consider `st_cast(x, "POINT")` to explode a multipoint into a row-per coordinate POINT data set". 

```R
leaflet() %>% addTiles() %>% addCircleMarkers(data = mpt)
Error in pointData.default(geometry) : 
  Don't know how to get location data from object of class sfc_MULTIPOINT,sfc
```

tim-salabim:
No problems, I never had any issues come up on my end, but I am currently not sure how we handle `MULTIPOINT` in mapview.
I second @mdsumner 's comment on a better warning!

mdsumner:
It's a minefield this stuff, I can't see any one of us keeping track of it, let alone someone just trying to use the tools!  It makes my head spin, and I mostly understand it.

bhaskarvk:
I am getting back to some development and I am getting headaches trying to read back all the code and figure out which piece goes where.

PR #360, opened 2017-01-22
add support for simple features MULTIPOLYGON

tim-salabim:
Fixes the issue that only the first outer polygon of a multipolygon object is rendered. Now all inner polygons are rendered as well so that holes appear correctly.


tim-salabim:
I was just made aware of @mdsumner s PR https://github.com/rstudio/leaflet/pull/354 so this is obsolete (as it's identical).

PR #364, opened 2017-01-25
Color function improvements

### LINKED ISSUE #300: colorFactor is broken with qualitative color brewer palette
For example try `previewColors(colorFactor("Set1", levels = 1:4), 1:4)`. Of the four colors shown, only one is actually part of `Set1`. It's unfortunate that the palette builder for quantitative values can't use a quantitative palette.

This is basically a duplicate of https://github.com/hadley/scales/issues/62 because the `colorFactor`code is derived from `scales::col_factor` and they share the same issue.

jcheng5:
This PR includes three significant quality-of-life improvements for users of the color palette functions: `colorNumeric`, `colorBin`, `colorQuantile`, and `colorFactor`.

### Native support for viridis palettes

E.g.: `pal <- colorNumeric("magma", values)`. Valid names are `"magma"`, `"inferno"`, `"plasma"`, and `"viridis"`.

### Allow reversing the color scale

E.g. `colorNumeric("magma", values, reverse=TRUE)`. So if your palette runs red-yellow-blue but you want the lower numbers to correspond to the blue end, not the red end, you can now just `reverse=TRUE` instead of retrieving the palette colors and reversing them manually.

### Fix #300: Don't interpolate qualitative palettes

When using qualitative RColorBrewer palettes with factors (or `colorBin` or `colorQuantile`), don't interpolate if possible (factor levels <= number of available colors). This is definitely the right behavior for `colorFactor`, I'm not positive whether it's the right behavior for `colorBin` and `colorQuantile`.

### DIFF
--- a/R/colors.R
+++ b/R/colors.R
@@ -301,7 +301,7 @@ toPaletteFunc <- function(pal, alpha, nlevels) {
 }
 
 # Strings are interpreted as color names, unless length is 1 and it's the name
-# of an RColorBrewer palette
+# of an RColorBrewer palette that is marked as qualitative
 toPaletteFunc.character <- function(pal, alpha, nlevels) {
   if (length(pal) == 1 && pal %in% row.names(RColorBrewer::brewer.pal.info)) {
     paletteInfo <- RColorBrewer::brewer.pal.info[pal,]
@@ -316,18 +316,13 @@ toPaletteFunc.character <- function(pal, alpha, nlevels) {
         }
       }
     }
-
-    return(scales::colour_ramp(colors, alpha = alpha))
+  } else if (length(pal) == 1 && pal %in% c("viridis", "magma", "inferno", "plasma")) {
+    colors <- viridis::viridis(n = 256, option = pal)
+  } else {
+    colors <- pal
   }
 
-  if (length(pal) == 1 && pal %in% c("viridis", "magma", "inferno", "plasma")) {
-    return(scales::colour_ramp(
-      viridis::viridis(n = 256, alpha = 1, option = pal),
-      alpha = alpha
-    ))
-  }
-
-  scales::colour_ramp(pal, alpha = alpha)
+  scales::colour_ramp(colors, alpha = alpha)
 }
 
 # Accept colorRamp style matrix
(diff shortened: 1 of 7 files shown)


R/colors.R
```diff
@@ -165,7 +175,7 @@ calcLevels <- function(x, ordered) {
   if (is.null(x)) {
     NULL
   } else if (is.factor(x)) {
-    levels(x)
+    as.character(levels(x))
```
hadley: `levels()` will always be a character

R/colors.R
```diff
@@ -210,30 +220,31 @@ colorFactor <- function(palette, domain, levels = NULL, ordered = FALSE,
...
-      if (any(is.na(x) != origNa)) {
-        warning("Some values were outside the color scale and will be treated as NA")
-      }
+    pf = safePaletteFunc(palette, na.color, alpha, nlevels = length(lvls) * ifelse(reverse, -1, 1))
+
+    origNa = is.na(x)
+    # Seems like we need to re-factor if hasFixedLevels, in case the x value
+    # has a different set of levels (like if droplevels was called in between
+    # when the domain was given and now)
+    x = factor(x, lvls)
```
hadley: I think calling `match()` directly here would be a clearer strategy

R/colors.R
```diff
@@ -268,35 +280,58 @@ colorFactor <- function(palette, domain, levels = NULL, ordered = FALSE,
...
+toPaletteFunc.character <- function(pal, alpha, nlevels) {
   if (length(pal) == 1 && pal %in% row.names(RColorBrewer::brewer.pal.info)) {
-    return(scales::colour_ramp(
-      RColorBrewer::brewer.pal(RColorBrewer::brewer.pal.info[pal, 'maxcolors'], pal),
-      alpha = alpha
-    ))
+    paletteInfo <- RColorBrewer::brewer.pal.info[pal,]
+    colors <- RColorBrewer::brewer.pal(paletteInfo$maxcolors, pal)
+
+    if (!is.null(nlevels) && paletteInfo$category == "qual") {
```
hadley: I feel like you could drop `paletteInfo$category == "qual"` - ideally you should use the colours as given whenever there are enough already (ie. you might have an ordered factor)

jcheng5:
@hadley This should be good now. I didn't realize RColorBrewer was doing anything more than interpolating its full color palettes to come up with smaller palettes. Now that I'm relying on its small-palette logic, the results look much, much better for all RColorBrewer palettes when used with `colorBin`, `colorQuantile`, and `colorFactor`.

Issue #355, opened 2017-01-10
leaflet sf support for linestring


timelyportfolio:
I am trying the new functionality with `LINESTRING` and getting an error. Most likely, this is a result of my ignorance, but

reproducible example

library(sf)
library(leaflet)

# make a line as shown in sf help
pts = matrix(1:10, , 2)
(ls1 = st_linestring(pts))

leaflet(ls1) %>% addPolygons()
error

Error in structure(as.data.frame(x), names = c("lng", "lat")) : 
  'names' attribute [2] must be the same length as the vector [1]
likely source of error

I believe the source of the error is in leaflet:::sf_coords.

leaflet:::sf_coords(ls1)
gives me

Error in as.data.frame.default(x) : 
  cannot coerce class "c("XY", "LINESTRING", "sfg")" to a data.frame
Naively, I tried leaflet(unclass(ls1)) %>% addPolygons() and it worked.

mdsumner:
@timelyportfolio just fyi some are classed over the base vector/matrix at the lowest level so need unclassing, see an approach m_or_v_XY in this stale mapview PR https://github.com/mdsumner/mapview/blob/simple-features-2/R/sf2.R.

I am not sure if leaflet ought to keep the Z or M by default if present.

tim-salabim:
To add to this, there are a couple more issues regarding sf support. For the 3 main sf classes for each feature type I see several errors:

```r
# devtools::install_github("rstudio/leaflet", force = TRUE)

library(leaflet)
library(sf)

### polygons --------------------------------------------------------------
poldata <- st_as_sf(mapview::gadmCHE)
class(poldata)
class(st_geometry(poldata))
class(st_geometry(poldata)[[1]])

leaflet() %>% addTiles() %>% addPolygons(data = poldata)
leaflet() %>% addTiles() %>% addPolygons(data = st_geometry(poldata))
leaflet() %>% addTiles() %>% addPolygons(data = st_geometry(poldata)[[1]]) # error

### lines -----------------------------------------------------------------
lindata <- st_as_sf(mapview::atlStorms2005)
class(lindata)
class(st_geometry(lindata))
class(st_geometry(lindata)[[1]])

leaflet() %>% addTiles() %>% addPolylines(data = lindata) # error
leaflet() %>% addTiles() %>% addPolylines(data = st_geometry(lindata)) # error
leaflet() %>% addTiles() %>% addPolylines(data = st_geometry(lindata)[[1]]) # error

[...]

PR #367, opened 2017-01-30
Support sf multipolygons

jcheng5:
This feature/refactor branch adds support for `MULTIPOLYGON` but also greatly streamlines the code needed to translate both sp and sf objects to Leaflet polygon data. See the giant comment in normalize.R, but the gist is that for any sp/sf object you only need to implement the most "natural" method among `to_ring`, `to_polygon`, `to_multipolygon`, or `to_multipolygon_list`, and rely on the default implementation of these methods to properly "box" the data the correct number of times (this was very tricky to get right previously).

`MULTIPOINT` is still not supported, but that's because I don't currently know how to implement the JS side without breaking clustering and crosstalk. At best we figure out a way to do this well, maybe by translating `MULTIPOINT` to `POINT` somewhere along the line; at worst we may need to add a new R function for adding this kind of data, that doesn't provide clustering or possibly even crosstalk.

@tim-salabim I am privately using some test cases that @timelyportfolio created using data from mapview. Do you mind if I copy the data into the leaflet package for unit testing? (I would rather copy the data than introduce a circular package dependency between mapview and leaflet)


bhaskarvk:
From leaflet.extras' POV, is there any changes I need to make or new functions I need to look out for?

Thanks

tim-salabim:
@jcheng5 sure, copy to your liking.

jcheng5:
@bhaskarvk If you haven't implemented a `polygonData` method, then I don't think so?

jcheng5:
Note: Fixes #355.

.travis.yml
```diff
@@ -1,11 +1,21 @@
...
 apt_packages:
-  - r-cran-rgdal
   - libproj-dev
 
+before_install:
+  - sudo add-apt-repository ppa:ubuntugis/ubuntugis-unstable --yes
+  - sudo apt-get --yes --force-yes update -qq
+  - sudo apt-get install --yes libudunits2-dev libproj-dev libgeos-dev libgdal-dev
+
+r_packages:
```
hadley: This would still be better in the DESCRIPTION

.travis.yml
```diff
@@ -1,11 +1,21 @@
 language: r
 sudo: required
 cache: packages
+dist: trusty
 
 apt_packages:
```
hadley: I think you can drop this given that you're installing `libproj-dev` below

R/normalize-sf.R
```diff
@@ -23,44 +23,46 @@ pointData.sfc_POINT <- function(obj) {
   )
 }
 
+#' @export
+pointData.POINT <- function(obj) {
+  check_crs(obj)
+
+  bbox <- sf_bbox(obj)
+
+  if (!is.matrix(obj)) {
```
hadley: When would this not be a matrix? A comment seems useful

R/normalize-sf.R
```diff
@@ -23,44 +23,46 @@ pointData.sfc_POINT <- function(obj) {
   )
 }
 
+#' @export
+pointData.POINT <- function(obj) {
+  check_crs(obj)
+
+  bbox <- sf_bbox(obj)
```
hadley: Just inline in `sf_bbox()`?

R/normalize.R
```diff
@@ -104,7 +104,7 @@ pointData <- function(obj) {
 #' @export
 pointData.default <- function(obj) {
   stop("Don't know how to get location data from object of class ",
-    class(obj)[[1]])
+    paste(class(obj), collapse = ","))
```
hadley: `", "`?

R/normalize.R
```diff
@@ -165,3 +165,80 @@ checkMatrix <- function(x) {
...
+# - to_ring
+#
+# Each of the specific sp/sf classes need only implement whichever ONE of those
+# actually makes sense (e.g. to_multipolygon_list.sfc,
+# to_multipolygon.MULTIPOLYGON, to_polygon.POLYGON, to_ring.LINESTRING). The
+# higher-level polygonData wrappers will simply call to_multipolygon_list(x),
+# and the default implementations of those methods will fall through to the next
+# level until a match is found.
+
+to_multipolygon_list <- function(x) {
```
hadley: If you want this to be extensible from outside leaflet, you'll need to export the generics too.

tim-salabim: exporting would highly be appreciated by selected developers that build their efforts on leaflet :-)

tim-salabim:
I am guessing their not in data because they are actually borrowed from mapview-package which has leaflet as a dependency. However, we could migrate them to leaflet and delete them from mapview if that is helping.

jcheng5:
@tim-salabim That's correct. I'm open to either, just let me know your preference.

.travis.yml
jcheng5: Sorry, can you elaborate? I'm pretty inexperienced when it comes to Travis. These packages are all in the DESCRIPTION under Suggests, but don't I need them listed here to get them to install for testing?

hadley: Any packages in DESCRIPTION are automatically installed

R/normalize-sf.R
jcheng5: I have to do this because I may turn it into a matrix below and that drops the bbox

tim-salabim:
My cunning plan for mapview is to focus my efforts on sf classes rather than sp. Would it make sense to have these sp data in leaflet and mapview provide sf versions of those?

.travis.yml
tim-salabim: I think sf (or some of its depends/imports, e.g. udunits2) specifically needs the dev version of these system libraries to be available and I don't think travis installs those by default.

hadley: @tim-salabim we're talking about R packages here, not system libraries

tim-salabim: @hadley I thought you're talking about the sudo apt-get calls, especially the one in line 12. If you're not, just ignore my comment

bhaskarvk: Agree with @hadley, you don't need the r_packages section here.

R/normalize.R
jcheng5: @tim-salabim You want to call it, or extend it? I don't think you want to call it--it returns very specific nested list structures that are only relevant to leaflet, most likely. And you don't want to implement it either--`derivePolygons` is what you want to implement. These `to_xxx` generics are just to make the implementation of `derivePolygons` less confusing for me, I don't think they make sense as part of the public contract of the package.

If you have strong feelings or a specific need that doesn't reflect the above, please let me know, I'm happy to reconsider.

tim-salabim: ping @timelyportfolio as this may be relevant for mapedit

timelyportfolio: @tim-salabim,@jcheng5 unexported should be fine.  I don't think we will need these in `mapedit`.  Thanks for all your work on this.  `sf` is definitely the future!

tim-salabim: @jcheng5 then no export needed. Cheers

PR #380, opened 2017-02-20
Move mapview data from extdata to data

jcheng5:
cc @tim-salabim


tim-salabim:
Cool, I will delete these from mapview asap.

jcheng5:
Do you want to wait until leaflet goes to CRAN? Hopefully will do the submission tomorrow.

tim-salabim:
good point. I guess these will simply be masked if loaded by both packages? 
And good to hear that leaflet release is close :-) Though that ups the pressure on my side...

jcheng5:
How much does it up the pressure on your side? There's not a compatibility problem between the github master of leaflet and the CRAN version of mapview, is there?

tim-salabim:
I haven't checked in a long time, but I think there are minor issues. E.g. `addMouseCoordinates` is likely to break. I would need to do a full check to be sure though

tim-salabim:
I've just tested mapview CRAN version with lealfet github version. 
As expected, `addMouseCoordinates` doesn't work anymore. For the same reason `latticeView` and `sync` won't behave properly. Both of these issues are already fixed in the development version on github. 
On the positive side, there's no errors breaking code execution on the R side. It's rather that things won't show up (properly) on the JS side. 

I don't think this is a biggy so I'd suggest you go ahead and submit to CRAN. I shouldn't be too far behind submitting to CRAN anyway, so I think the damage will be minimal. Plus, the new features of leaflet surely outweigh the damage.

PR #385, opened 2017-02-23
Ability to show/hide layer groups based on zoom level

jcheng5:
Adds a `groupOptions` function that currently only lets you set one option: `zoomLevels`.

Example: http://rpubs.com/jcheng/leaflet-multiscale


tiernanmartin:
In the example you provided, it seems that the tiles disappear when the zoom level is within the set range - is that by design?

leaflet-ss-zoomgroups

jcheng5:
It's not that the tiles disappear, they're replaced by Stamen.TonerLite which is pretty plain at that zoom level. Sorry, bad choice.

mfincker:
The zoomLevel option does not work when you supply a single value instead of a range.

```
library(leaflet)

pal <- colorQuantile("YlOrRd", quakes$mag)

leaflet() %>%
  # Basic markers
  addTiles(group = "basic") %>%
  addMarkers(data = quakes, group = "basic") %>%
  # When zoomed in, we'll show circles at the base of each marker whose
  # radius and color reflect the magnitude
  addProviderTiles(providers$Stamen.TonerLite, group = "detail") %>%
  addCircleMarkers(data = quakes, group = "detail", fillOpacity = 0.5,
    radius = ~mag * 5, color = ~pal(mag), stroke = FALSE) %>%
  # Set the detail group to only appear when zoomed in
  groupOptions("detail", zoomLevels = 7)
```

screen shot 2017-02-23 at 9 10 28 pm

jcheng5:
Thanks @mfincker, should be fixed now!

jcheng5:
@bhaskarvk Do you feel like reviewing this? No pressure if you're too busy.

bhaskarvk:
I will over the weekend. Or do you need it sooner than that ?

jcheng5:
@bhaskarvk That'd be great. Thanks!

inst/htmlwidgets/leaflet.js
```diff
@@ -534,6 +558,7 @@ _htmlwidgets2.default.widget({
         }
 
         preventUnintendedZoomOnScroll(map);
+        setupShowHideGroupsOnZoom(map);
```
bhaskarvk: One thing I am a bit concerned about is, that this sets up the `zoomend` event listener, no matter whether any groups were added with zoomLevels or not.
Can't this whole thing be setup from `methods.setGroupOptions`, so that the event listener is registered only when someone sets up groupOptions?

Other than that looks good.

jcheng5:
@bhaskarvk Thanks, I made the change you suggested.

PR #391, opened 2017-02-27
addPolygons() now responds to crosstalk events

cpsievert:
For example,

```r
library(crosstalk)
library(sf)
library(plotly)

nc <- st_read(system.file("shape/nc.shp", package="sf")) %>% 
  st_transform(4326)

shared_nc <- nc %>% SharedData$new()

bscols(leaflet(data = shared_nc) %>%
         addTiles() %>%
         addPolygons(opacity = 1,
                     color = 'white',
                     weight = .25,
                     fillOpacity = .5,
                     fillColor = 'blue',
                     smoothFactor = 0),
       plot_ly(shared_nc, x = ~BIR74, y=  ~SID79, type = 'scatter', mode = 'markers')       
)
```

leaf

TODO:

  - [ ] Implement `addPolylines()` etc?
  - [ ] Should 2D brush events on polygons emit a crosstalk event?

### DIFF
--- a/R/layers.R
+++ b/R/layers.R
@@ -1087,8 +1087,11 @@ addPolygons <- function(
     dashArray = dashArray, smoothFactor = smoothFactor, noClip = noClip
   ))
   pgons = derivePolygons(data, lng, lat, missing(lng), missing(lat), "addPolygons")
-  invokeMethod(map, data, 'addPolygons', pgons, layerId, group, options, popup, popupOptions, safeLabel(label, data), labelOptions, highlightOptions) %>%
-    expandLimitsBbox(pgons)
+  invokeMethod(
+    map, data, 'addPolygons', pgons, layerId, group, options, popup,
+    popupOptions, safeLabel(label, data), labelOptions, highlightOptions,
+    getCrosstalkOptions(data)
+  ) %>% expandLimitsBbox(pgons)
 }
 
 #' @rdname remove
--- a/inst/htmlwidgets/leaflet.js
+++ b/inst/htmlwidgets/leaflet.js
@@ -1596,9 +1596,9 @@ methods.addRectangles = function (lat1, lng1, lat2, lng2, layerId, group, option
  * @param lat Array of arrays of latitude coordinates for polygons
  * @param lng Array of arrays of longitude coordinates for polygons
  */
-methods.addPolygons = function (polygons, layerId, group, options, popup, popupOptions, label, labelOptions, highlightOptions) {
+methods.addPolygons = function (polygons, layerId, group, options, popup, popupOptions, label, labelOptions, highlightOptions, crosstalkOptions) {
   if (polygons.length > 0) {
-    var df = new _dataframe2.default().col("shapes", polygons).col("layerId", layerId).col("group", group).col("popup", popup).col("popupOptions", popupOptions).col("label", label).col("labelOptions", labelOptions).col("highlightOptions", highlightOptions).cbind(options);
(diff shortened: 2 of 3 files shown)


bhaskarvk:
Nice! So in theory this should work for all shapes like rectangles, circles, circleMarkers, polylines, right ? provided code for each of these shapes is changed accordingly.

jcheng5:
I was originally thinking polygon brushing would be really hard to implement, because it'd either have to work by 1) any intersection between the brushing rectangle and the polygon, or 2) complete coverage of the polygon by the brushing rectangle.

But a nice easier-to-implement alternative would be 3) the brushing rectangle includes the center-of-mass of the polygon (is that the right term for it?). I believe there are fast routines to do this in R?

jcheng5:
I mean, obviously the correct term is "centroid" not center-of-mass, and obviously it's just the arithmetic mean of all of the points in the polygon. :)

cpsievert:
Yea, that sounds reasonable, and maybe we could leverage `sf::st_centroid()`?

bhaskarvk:
For irregular polygons centroids can fall out of the polygon/s. 
You may also want to evaluate https://www.mapbox.com/blog/polygon-center/. It is a JS solution, so in theory it should be easy to integrate.

jcheng5:
The mapbox solution is really interesting. I can see how it would be much better than centroid for labeling but I'd want to try out both for brushing to see how they feel.

bhaskarvk:
Yup agreed.

mdsumner:
We can decompose to triangles for finer elements, as illustrated in my sfdct package. License is a problem but in time we might get CGAL or another alternative.

mdsumner:
Also fwiw this is what I was going after here: https://twitter.com/mdsumner/status/830380049815597056

krlmlr:
We could make very good use of this pull request. Is there anything we can help with so that it can be accepted?

cpsievert:
@krlmlr this PR only implements the ability for leaflet polygons to _respond_ to crosstalk events. I'm pretty sure @jcheng5 hesitated to take this on because ideally, when this is added, you'd also have the ability to brush polygons to _emit_ crosstalk events (which is what most of the comments are about). If you wanted to take a stab at implementing that, I'm pretty sure that'd increase the chances of this landing

jcheng5:
Yeah that would help a lot. Even “just” using the centroid algorithm.

tim-salabim:
FWIW, @timelyportfolio came up with [this](https://github.com/r-spatial/mapedit/blob/master/experiments/select_crosstalk.R) some time ago to enable this kind of thing. Maybe that could serve as a basis for implementation.

In action here:
https://twitter.com/timelyportfolio/status/942519152731926529

krlmlr:
Responding to crosstalk events is good enough for our use case. I intend to respond to `$map_shape_click` events and update the selection manually via `shared_data$selection()` . I don't see the value of brushing polygons for our use case.

I think it would be useful to enable this for all other shape types, we could file an issue to implement brushing at some point if necessary. I haven't found a documentation of this limitation -- why we can crosstalk points but not polygons -- where can we document it?

krlmlr:
I'm also missing the ability to select/deselect individual objects in the crosstalk framework, this doesn't seem to work neither with leaflet nor with plotly. Has this been considered?

sstoltzman-rclco:
Hello, I have found quite a few projects that would benefit from this. It works really well but it is not in the main rstudio/leaflet branch. Is there any reason there's a hangup in merging? I am happy to help if I can be pointed in the right direction.

Issue #402, opened 2017-03-19
label formula fails when using crosstalk


kent37:
Labels can't be specified as formulas when the map data is a crosstalk SharedData. Simple example:

```
library(crosstalk)
library(leaflet)

shared_quakes <- SharedData$new(quakes[sample(nrow(quakes), 100),])
map = leaflet(shared_quakes,) %>% addTiles() %>% 
  addMarkers(label=~as.character(depth))
```

gives error

```
Error in UseMethod("metaData") : 
  no applicable method for 'metaData' applied to an object of class "c('SharedData', 'R6')"
```

The problem is a simple mis-spelling at line 2 of `normalize-SharedData.R`. `metaData.sharedData` should be `metaData.SharedData`.

kent37:
There are two additional complications to this.
- The result of calling `metaData.SharedData` may itself need to be unpacked with another call to `metaData`, for example if the `SharedData` is wrapping a `Spatial*` object. This is easily fixed by adding another call to `metaData`. `pointData.SharedData` and `polygonData.SharedData` already do something similar for the same reason.
- The second issue is deeper and arises because the `leaflet` and `crosstalk` packages have different ways to extract data from an object. `leaflet` uses (and extends) `metaData` while `crosstalk` uses `lazyeval::find_data`. Since `crosstalk` is not aware of `leaflet::metaData`, it does not benefit from the definitions in `leaflet`.

Here is an example showing both problems:

```r
library(crosstalk)
library(leaflet)
library(sp)

# Get a SpatialPointsDataFrame
data(meuse)
coordinates(meuse) <- ~x+y
proj4string(meuse) <- CRS("+init=epsg:28992")
meuse =  spTransform(meuse, CRS("+init=EPSG:4326"))
meuse_shared = SharedData$new(meuse)

# Fails because filter_select doesn't know how to get data from a SpatialPointsDataFrame
filter_select('landuse', 'Land Use', meuse_shared, ~landuse)
# Error: Do not know how to find data associated with `x`

# Fails because metaData.SharedData returns a SpatialPointsDataFrame, not a data.frame
leaflet(meuse_shared) %>% addTiles() %>% addMarkers(label=~landuse)
# Error in as.character(x) :
#  cannot coerce type 'closure' to vector of type 'character'
```

[...]

PR #415, opened 2017-04-05
More formula fixes for #402

kent37:
Fixes both issues shown in the second code snippet in #402:

 - `filter_select` with formula fails because it doesn't know how to get data from a `SpatialPointsDataFrame`
- `leaflet::addMarkers` and friends with formula for `label` fail because `metaData.SharedData` returns a `SpatialPointsDataFrame`, not a `data.frame`


kent37:
The `metaData.SharedData` problem is fixed in 5a6ef86060baad60afbd6824fe62d3c8f8c9bb14 and the `find_data` problem is complicated; see #402 for a workaround.

PR #429, opened 2017-05-23
Fixed handling non-finite values in colorBin arg domain.

CreMindES:
After investigating the special snowflake error message: "Error in if (bins < 2) { : argument is of length zero", it turned out that colorBin does not check for non-finite values in its 'domain' argument.

### DIFF
--- a/R/colors.R
+++ b/R/colors.R
@@ -112,6 +112,8 @@ colorBin <- function(palette, domain, bins = 7, pretty = TRUE,
   # breaks are specified
   if (missing(domain) && length(bins) > 1) {
     domain = NULL
+  } else if (!(length(domain) == 2 && all(is.finite(domain)))) {
+    stop ("'Domain' must be of length 2 and contain finite values only!")
   }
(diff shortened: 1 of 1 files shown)


CreMindES:
Apologies, after reviewing the contents of test-colors.R, I need additional information about the design choices of domain.

I can think of two common approaches:
1. supply explicit bounds e.g. c(1,100)
2. supply the whole vector and let the function find out its extent

Both of them has their pros and cons, and of course we always try to respect backward compability.
May I ask you to elaborate on the idea of domain in case of NULL or single number?

jcheng5:
Can you give an example of the case you're trying to fix?

CreMindES:
The code below gives that not so nice error message, due to not handled NAs.

`bw <- c("black", "white");
binpal <- colorBin(
        palette =  bw,
        domain = c( NA, NA ),
        bins = 100, 
        pretty = T,
        alpha = 0.5,
        na.color = "red" );
binpal(1);`

> Error in if (bins < 2) { : argument is of length zero

PR #442, opened 2017-07-18
Auto show/hide legends with support for LayerControl and groupOptions

bhaskarvk:
@jcheng5 Now this PR supports overlay groups as well as groups shown/hidden via the `groupOptions` method. The only thing I can't do is show/hide legend tied to a base layer, coz in that case we only get baselayerchange event, while for properly tying I need to know when layer got added and which one got removed. It can be hacked if needed, but I am not convinced of a use case where you need to auto show/hide legends at basemap level.

Demo for the groupOptions at https://rpubs.com/bhaskarvk/group-options (zoom to level 7 to show the legend , zoom out to hide).

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -12,5 +12,4 @@
 ^javascript$
 ^package\.json$
 ^node_modules$
-
 ^data-raw$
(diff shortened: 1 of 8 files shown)


osh-chicago:
This is great - thank you so, so, so much @bhaskarvk! Please let us know when it's available at devtools::install_github('rstudio/leaflet').

bhaskarvk:
That's up to Joe ;)

bhaskarvk:
in the mean time `devtools::install_github('bhaskarvk/leaflet', ref='feature/attachLegend')` should get you that functionality.

jcheng5:
Thanks @bhaskarvk for another great contribution :) And sorry for the delay, everyone.

PR #449, opened 2017-08-03
Fix typos in docs

jameslamb:
This PR addresses a few small typos in documentation. The mis-placed `)` in the documentation for `icons` was causing an `R CMD check` note when using `#' @inheritDotParams leaflet::addPolygons` in other packages.


jcheng5:
Thanks!

Issue #491, opened 2018-02-09
addLegend should support `data` argument


jcheng5:
All the other `addXXX` functions support an explicit `data` argument, while `addLegend` assumes that your leaflet object has it. There's no reason for this, just an oversight that should be trivial to fix.

jcheng5:
See https://github.com/rstudio/leaflet/issues/485

schloerke:
Adding also to 
* addTiles
* addRasterImage
* addWMSTiles
* addGeoJSON
* addLayersControl

schloerke:
fixed in https://github.com/schloerke/leaflet/pull/1

Issue #477, opened 2017-11-15
addLegend group not working when group is in addLayersControl baseGroups


wimpel:
In my shiny-app, i've got a leaflet, based on reactive userinput

```
  output$leafletMap2 <- renderLeaflet({
    leaflet(df[which(!is.na(df$latitude)), ]) %>% addTiles() %>% setView(4.213, 51.876, zoom = 7)
  })

  observe({
    leafletProxy("leafletMap2", data = filteredData() ) %>% 
      clearGroup("test1") %>%
      clearGroup("test2") %>%
      clearGroup("test3") %>%
      addCircleMarkers(lng = ~longitude,
                       lat = ~latitude,
                       radius = 6,
                       color = ~pal(process),
                       stroke = FALSE,
                       fillOpacity = 1,
                       popup = paste(filteredData()$time, "<br>", filteredData()$description),
                       popupOptions = popupOptions(closeOnClick = FALSE),
                       group = "test1") %>%
      addCircleMarkers(lng = ~longitude,
                       lat = ~latitude,
                       radius = 6,                        
                       color = ~pal(PROCES),
                       stroke = FALSE,
                       fillOpacity = 1,
                       clusterOptions = markerClusterOptions() ,
                       popup = paste(filteredData()$time, "<br>", filteredData()$description),
                       popupOptions = popupOptions(closeOnClick = FALSE),
                       group = "test2") %>%
      addHeatmap(lng = ~longitude,
                 lat = ~latitude,
                 blur = 15,                 
                 max = 1, 
                 radius = 10,              
                 intensity = 0.25,       
                 group = "test3") %>%
      clearControls() %>%
      addLegend(position = "bottomleft",
                pal = pal,
                values = sort( unique(filteredData()$PROCES) ),
                opacity = 1,
                group = "test1") %>%
      addLayersControl(baseGroups = c("test3", "test1", "test2"), 
                       #overlayGroups = c("test3", "test1", "test2"), 
                       options = layersControlOptions(collapsed = FALSE)
      )
  })
```
I would like the legend to show **only** when test3 is the selected layer. This does not work when I select a layer from the baseGroups (with radiobuttons). However, it does work when I add the laters to the overlayGroups (with checkboxes). I rather do not want to do that, because I want the user not to be able to select multuiple layers at once).

Not sure if this is a bug-report, feature request or that I'm simply missing somthing ;-).

bhaskarvk:
[...]

PR #462, opened 2017-09-09
add method argument to addRasterImage to enable neighrest neighbor interpolation when projecting categorical rasters

tim-salabim:
This should fix https://github.com/rstudio/leaflet/issues/461 and https://github.com/rstudio/leaflet/issues/219.

I also took the liberty to address https://github.com/rstudio/leaflet/issues/411

PR #458, opened 2017-08-30
update markercluster->v0.1.1, leaflet-measure->2.1.7, leaflet-omnivore->0.3.4

timelyportfolio:
I updated 

- leaflet-markerCluster -  All seems to work except for `freezeAtZoom` using [example](https://github.com/rstudio/leaflet/blob/master/inst/examples/marker-clustering.R#L17-L18)
- leaflet-measure - think all good
- leaflet-omnivore - tried and doesn't seem to work
    ```
    topoData <- readLines("https://d3js.org/us-10m.v1.json") %>% paste(collapse = "\n")
    
    leaflet() %>% setView(lng = -98.583, lat = 39.833, zoom = 3) %>%
      addTiles() %>%
      addTopoJSON(topoData, weight = 1, color = "#444444", fill = FALSE)
    ```

PR #492, opened 2018-02-12
v2.0.0

schloerke:
Resolves #476 (which merges #453, #458, #462, #405, #396, #393).

Remaining PR's to possibly include:
- * #342 - horizontal legend
  * lots of js changes
  * code review
  * test against all legend types
  - [x] CLA Signed
  * Needs more work.  Will wait for next release
* #346 - crosstalk for plotly
  * Wait until next release
* #391 - crosstalk for addPolygons()
  * Wait until next release.  Would like to address this functionality properly
- [x] #388 - add parameter right to color cutting
  - [x] CLA Signed
- [x] #450 - map center info to doc
  - [ ] CLA Signed. Only minor documentation. Merged
* #477 - add legend group not working when group is in addLayersControl baseGroups
  * scope this out
- [x] #497 - Removed deprecated br clear
  - [x] CLA Signed

Remaining Issues to address:
* [x] #498 documenting where .json files are located

Tasks left before merging

- [x] update pkg version
  * ~~1.2.0~~
  * 2.0.0
    * with the upgrade to 1.3.1, there will most likely be breaks in the js code that people submit.  We should make it a 'major' version update.
- [x] Should we update to [leaflet.js 1.3.1](https://github.com/Leaflet/Leaflet/blob/master/CHANGELOG.md) ? 
  * ~~no.  release to public.  then upgrade to 1.3.1 (or latest) in next release~~
  * Yes! leaflet.extras and leaflet.esri need updating.  Since we are doing work for both packages, then we should update leaflet version as well.
- [x] Packages ready for rstudio/leaflet 2.0.0
  - [x] leaflet.extras 
    * needs leaflet 1.3.1 updates
  - [x] leaflet.esri
    * needs leaflet 1.3.1 updates
  - [x] mapview
  - [x] mapedit
  - [x] sf
- [x] Internal lints
  * with the update of leaflet version, maybe this could be included in the release
- [x] issue #491 - addLegend should support `data` arg
- [x] leaflet.extras has no zIndex = NULL 
- [x] update NEWS with information
  * ... ongoing process
- [x] raster auto method, and color
- [x] add auto method for raster data
- [ ] add raster examples
  - [ ] numeric
  - [ ] character
  - [ ] factor
  - [ ] weird
- [x] Fix known broken examples (added below)
- [x] run `revdepcheck::revdep_check()` on final version
 

Known broken examples:

* FIXED BELOW - From #458 (which might be related to an updated leaflet-omnivore)
```r
topoData <- readLines("https://d3js.org/us-10m.v1.json") %>% paste(collapse = "\n")
leaflet() %>% setView(lng = -98.583, lat = 39.833, zoom = 3) %>%
  addTiles() %>%
  addTopoJSON(topoData, weight = 1, color = "#444444", fill = FALSE)
# map shows up with no topo viz
```


jcheng5:
@schloerke Your new commits look good to me. Let's discuss the remaining PR's over vchat--grab some time on my calendar today or tomorrow?

timelyportfolio:
@schloerke so excited for this.  Please let me know how I can help.  Thanks!

tim-salabim:
@schloerke I have a few points I'd like to make before release:

1. I think it is possible in 1.x to specifically set the pane for many layer types (see https://github.com/rstudio/leaflet/issues/427). My JS knowledge is not good enough to test this, so I would appreciate if you could look into this option (and potentially include it in the `add*` functions). We have had a few enquiries if this can be done, especially if a) it would be possible to render `rasterImage` on top of a vector layer and b) if it is possible to specify a strict order for vector layers so that switching them on and off prevents rending of polygons on top of circleMarkers/Polylines.
2. One thing that people frequently ask is rendering of `rasterImage` layers that don't have a coordinate reference system. See e.g. https://github.com/r-spatial/mapedit/issues/67 Would it be possible to add support for this in the next release?
3. The label formatting seems to have changed from 0.7 to 1.x. Especially, the label is now displayed at the center of a feature by default (documented on the leafletjs side). I wonder if it makes sense for the leaflet package to change this default behaviour back to what it used to be? I know that it is not desirable to mess with the upstream defaults, but feel like this is worthy an exception as it simply feels weird and kind of ugly with the new default. Don't get me wrong, this is just to raise this point, I have no problem at all with sticking to the JS default as I can easily adjust it in mapview/mapedit (see code below). 
```r
library(leaflet)

### default
leaflet() %>%
  addTiles() %>%
  addPolygons(data = gadmCHE, label = ~NAME_1)

### 'old' behaviour following the cursor
leaflet() %>%
  addTiles() %>%
  addPolygons(data = gadmCHE, label = ~NAME_1, 
              labelOptions = labelOptions(sticky = TRUE))
```
Also, would it be possible to add labels to circleMarkers? So far they are not available for this feature type.

Thanks in advance. I am really happy that you have taken on this  work and look forward to result!
FWIW, I have created a `leaf1x` branch in mapview.

timelyportfolio:
@bhaskarvk, @jcheng5 Does anyone remember what happened with #303 (see #170 and #496)?  I see it was closed with no comment, and I do not see any other implementation off this feature.  Thanks!

jcheng5:
The `restyle` operation seems like something you should be able to do to anything, not just GeoJSON?

timelyportfolio:
@jcheng5 based on the Leaflet docs, `setStyle` is available for

[`path`](http://leafletjs.com/reference-1.3.0.html#path-setstyle)

[`FeatureGroup`](http://leafletjs.com/reference-1.3.0.html#featuregroup-setstyle)

[`GeoJSON`](http://leafletjs.com/reference-1.3.0.html#geojson-setstyle)

... but this list might not be comprehensive.

schloerke:
Just to record somewhere... 

The +/- and layer control (and most likely others) may appear larger as leaflet is adding different css classes than in 0.7.x.  Mainly the `.leaflet-touch` class is being added to my personal chrome browser which alters which css values are used.  This class is added according to the reasoning below.

```js
// @property touch: Boolean
// `true` for all browsers supporting [touch events](https://developer.mozilla.org/docs/Web/API/Touch_events).
// This does not necessarily mean that the browser is running in a computer with
// a touchscreen, it only means that the browser is capable of understanding
// touch events.
```

Icon locations were not properly displayed before.  The anchor was top aligned by default, not center aligned, like it should be.  This was caused by a R to js translation error.  rstudio/leaflet v1.1.0 behavior was incorrect.

schloerke:
Known broken example:
```r
topoData <- readLines("https://d3js.org/us-10m.v1.json") %>% paste(collapse = "\n")

leaflet() %>% 
  setView(lng = -98.583, lat = 39.833, zoom = 3) %>%
  addTiles() %>%
  addTopoJSON(topoData, weight = 1, color = "#444444", fill = FALSE)
```

is fixed by changing the data source:
```r
topoData <- readLines("http://bl.ocks.org/mbostock/raw/4090846/us.json") %>% paste(collapse = "\n")

leaflet() %>% 
  setView(lng = -98.583, lat = 39.833, zoom = 3) %>%
  addTiles() %>%
  addTopoJSON(topoData, weight = 1, color = "#444444", fill = FALSE)
```

I believe Bostock is projecting the data.  This raw data from an old block is not geo projected.  The raw data is leaflet ready, where as the original data is ready for D3.

timelyportfolio:
@schloerke thanks for incredibly hard work on this.

jrisi256:
@jcheng5 Hi! I submitted issue #496 (related to #152, #163, #170, and #303), and I didn't quite understand your response to @timelyportfolio by saying "The restyle operation seems like something you should be able to do to anything, not just GeoJSON?" Could you explain further if there are any plans to implement setStyle or restyle? Or if not, could you point out a way/gesture towards how to implement it using some Javascript? Thanks for your time and for helping make such a great package.

PR #501, opened 2018-03-21
Add options parameter to fitBounds

mz8i:
The fitBounds method in leaflet.js allows, like many other methods, to pass an options object. The R interface didn't have that possibility. After the change, it is possible to call for example.

```R
leaflet() %>% fitBounds(1, 1, 2, 2, options = list(animate=TRUE))
```

Wasn't sure if the bindings code in ```/inst/legacy/www/leaflet.js ``` should be changed, however...

### DIFF
--- a/R/methods.R
+++ b/R/methods.R
@@ -35,6 +35,8 @@ setView <- function(map, lng, lat, zoom, options = list()) {
 
 #' @describeIn map-methods Set the bounds of a map
 #' @param lng1,lat1,lng2,lat2 the coordinates of the map bounds
+#' @param options a list of zoom/pan options (see
+#'   \url{http://leafletjs.com/reference.html#zoom/pan-options})
 #' @export
 fitBounds <- function(map, lng1, lat1, lng2, lat2, options = list()) {
   bounds = evalFormula(list(lat1, lng1, lat2, lng2, options), getMapData(map))
--- a/R/methods.R
+++ b/R/methods.R
@@ -36,8 +36,8 @@ setView <- function(map, lng, lat, zoom, options = list()) {
 #' @describeIn map-methods Set the bounds of a map
 #' @param lng1,lat1,lng2,lat2 the coordinates of the map bounds
 #' @export
-fitBounds <- function(map, lng1, lat1, lng2, lat2) {
-  bounds = evalFormula(list(lat1, lng1, lat2, lng2), getMapData(map))
+fitBounds <- function(map, lng1, lat1, lng2, lat2, options = list()) {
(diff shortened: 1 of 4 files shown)


schloerke:
Hi @mz8i  

Could you ...

"
 2. Ensure that you have signed the [individual](https://rstudioblog.files.wordpress.com/2017/05/rstudio_individual_contributor_agreement.pdf) or [corporate](https://rstudioblog.files.wordpress.com/2017/05/rstudio_corporate_contributor_agreement.pdf) contributor agreement as appropriate. You can send the signed copy to jj@rstudio.com.
"
Excerpt from [rstudio/httpuv package](https://github.com/rstudio/httpuv/blob/master/CONTRIBUTING.md), which you had no reason to know existed.

Please let me know when you've sent the email.  Thank you for your help!

Best,
Barret

mz8i:
Hi @schloerke sure, I will.
By the way, from what I can see my change in this PR has been also covered by your recent work on moving to leaflet 2.0. The only difference is that it seems my code has the proper docstrings, while the current master is missing a description for the "options" param for method.R -> fitBounds()
Feel free to not merge this PR and just amend the docstrings if that's easier

schloerke:
@mz8i Thanks.  Let me know when you've submitted and I'll merge your other PR.  (The contributor agreement will still work for any other PRs you submit.)

Ah! Thank you for spotting that. This line shows it should point to the leaflet.js docs. Kinda wonky with the roxygen `@describeIn map-methods` tag allowing docs from other locations to describe immediately located parameters.
https://github.com/rstudio/leaflet/blob/163ea8e340f18a6d8b6eca35f5893c4b9d6e1bc7/R/methods.R#L8-L12

https://github.com/rstudio/leaflet/blob/163ea8e340f18a6d8b6eca35f5893c4b9d6e1bc7/R/methods.R#L52-L55

Closing for no action needed.

mz8i:
@schloerke oh, I didn't spot that fragment at the beginning - so does it mean that all parameters listed at the top of the file apply to every method? And the ones described directly above methods are just the ones specific for each method. Out of curiosity, I see the version of the leaflet reference linked is 1.3.1 - is that correct?

schloerke:
Yes, the leaflet.js version is now 1.3.1.  Any docs referencing a 1.3.1 url get redirected to 1.3.0 while maintaining the url hash.

The documentation is all under a single rd file by using the roxygen tag `@describeIn`.  R does not like it when parameters are documented twice within the same file.  roxygen keeps top-down order from the documentation within the file.  Yup, all specific parameters are added for each function.

schloerke:
Closing, as PR is duplicated in current functionality

Issue #463, opened 2017-09-12
Canvas renderer?


araichev:
Hi folks,

Is there a way to use [Leaflet canvas rendering](http://leafletjs.com/reference-1.2.0.html#canvas)?  I have a big GeoJSON file i'd like to map, and using canvas instead of SVG would help.

bhaskarvk:
We are not on leaflet js 1.x, we are still using leaflet js 0.7.7 and according to http://leafletjs.com/reference-0.7.7.html#path-canvas , canvas render of 0.7.x suffers from a performance bug. We are soon going to migrate to 1.x and we will add an option to the `leafletOptions` function to support canvas renderer.

araichev:
Thanks for the information.

cmcaine:
`leafletOptions(preferCanvas = TRUE)` works as expected in v2 but isn't documented. Is it still recommended against?

schloerke:
@cmcaine I do not see any warnings in the latest leaflet.js documentation about canvas (http://leafletjs.com/reference-1.3.0.html#canvas) as I do in the [prior link](http://leafletjs.com/reference-0.7.7.html#path-canvas).

PR #521, opened 2018-04-10
add note about preferCanvas in leafletOptions()

schloerke:
Fixes https://github.com/rstudio/leaflet/issues/463#issuecomment-380124722

```
Added `preferCanvas = FALSE` argument to `leafletOptions()` (#521)
```

```
library(leaflet)
# right click on polygon, should be a 'regular' right click (SVG) (back, forward, reload, print, etc.)
leaflet() %>%
  addTiles() %>%
  addPolygons(data = gadmCHE, label = ~NAME_1)

# right click on polygon, should ask about saving an image or copying an image
leaflet(options = leafletOptions(preferCanvas = TRUE)) %>%
  addTiles() %>%
  addPolygons(data = gadmCHE, label = ~NAME_1)

# right click on polygon, should be a 'regular' right click (SVG)
leaflet(options = leafletOptions(preferCanvas = FALSE)) %>%
  addTiles() %>%
  addPolygons(data = gadmCHE, label = ~NAME_1)
```

PR task list:
- [x] Update NEWS

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -64,6 +64,9 @@ BUG FIXES AND FEATURES
 
 * Added `data` parameter to remaining `addXXX()` methods, including addLegend. (f273edd, #491, #485)
 
+* Added `preferCanvas = FALSE` argument to `leafletOptions()` (#521)
+
+
 leaflet 1.1.0
 --------------------------------------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


cmcaine:
Maybe the default should be NULL (use whatever leaflet.js prefers)?

Same result with current version of leaflet, but could conceivably change.

schloerke:
Leaflet has the default value to be `FALSE`.  Will merge shortly

schloerke:
@cmcaine You are correct. Missed it earlier.  Updating to `NULL` in #523

PR #522, opened 2018-04-12
allow all pointer events on popup tip container

### LINKED ISSUE #519: popups don't close on click anymore
Hi,
more testing with 2.0.0 reveals that popups are not closed when clicking the same location for the second time (anywhere else on the map closes the popup or opens another one). This used to be the default behaviour and I have come to like it. Thus, I would like to keep this behaviour in mapview, however, I cannot figure out how to manipulate `popupOptions` to bring it back.

Am I missing something obvious?

```r
library(leaflet)

leaflet() %>%
  addTiles() %>%
  addPolygons(data = gadmCHE, 
              popup = ~NAME_1)

leaflet(options = leafletOptions(closePopupOnClick = TRUE)) %>%
  addTiles() %>%
  addPolygons(data = gadmCHE, 
              popup = ~NAME_1,
              popupOptions = popupOptions(closeOnClick = TRUE))
```

schloerke:
Fixes #519

`pointer-events` were blocked in css. Unblocked by setting all `pointer-events` to `all`.

Added cursor shape to match regular map interaction

```r
library(leaflet)

leaflet() %>%
  addTiles() %>%
  addPolygons(data = gadmCHE, 
              popup = ~NAME_1)
```

### DIFF
--- a/inst/htmlwidgets/lib/rstudio_leaflet/rstudio_leaflet.css
+++ b/inst/htmlwidgets/lib/rstudio_leaflet/rstudio_leaflet.css
@@ -18,3 +18,10 @@
   border-right: 6px solid transparent;
   /* right: -16px; */
 }
+
+.leaflet-popup-pane .leaflet-popup-tip-container {
+  /* when the tooltip container is clicked, it is closed */
+  pointer-events: all;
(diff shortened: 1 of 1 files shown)


schloerke:
For full reference.... https://www.w3.org/TR/SVG11/interact.html#PointerEventsProperty

PR #544, opened 2018-05-11
use unpkg for all non http marker locations

schloerke:
The leaflet cdn does not support https anymore.  Loading icons fails on https pages now.  Using unpkg.com instead.

```r
leaflet() %>% addTiles() %>% addMarkers(0,0)
```


schloerke:
The addition of the extra doc/libs files is due to a correction in a htmlwidgets bug.  The files were suppose to be there in the first place.  (For future, we should prune and ship only necessary files.)

PR #563, opened 2018-06-25
fix broken leaflet-search mouse events within shiny

schloerke:
Search caused the latlng values to be poisoned with a layer attribute.  When the latlng value is serialized for json in method.js, this is not possible due to recursion.  This fix trims all latlng values for mouse handlers for shiny.

Original Issue: https://github.com/bhaskarvk/leaflet.extras/issues/104#issuecomment-399845514

```r
library(shiny)
library(leaflet)
library(leaflet.extras)

ui <- fluidPage(
  leafletOutput("map")
)

  data <- data.frame(lat = 20:22, lng = 30:32, name = c("first", "second", "third"))
  
  output$map <- renderLeaflet({
    leaflet(data) %>%
      addTiles() %>%
      addMarkers(lng = ~lng, lat = ~lat, label = ~name, group = "names") %>%
      addSearchFeatures(targetGroups = "names")
  })
  
  observe({
    input$map_marker_click
    cat('clicked\n')
  })
}

shinyApp(ui, server)
```

PR task list:
- [x] Update NEWS

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -3,7 +3,8 @@ leaflet 2.0.1.9000
 
 BUG FIXES and IMPROVEMENTS
 * Require viridis >= 0.5.1 to avoid namespace issues with viridisLite (#557)
-* Fixed recursive JSON object serialization issue for mouse events within shiny (#563)
+* Fixed broken mouse events after using leaflet-search from leaflet.extras within shiny applications (#563)
+
 
 
 leaflet 2.0.1
--- a/NEWS
+++ b/NEWS
@@ -3,6 +3,8 @@ leaflet 2.0.1.9000
 
 BUG FIXES and IMPROVEMENTS
 * Require viridis >= 0.5.1 to avoid namespace issues with viridisLite (#557)
+* Fixed recursive JSON object serialization issue for mouse events within shiny (#563)
+
 
 leaflet 2.0.1
 --------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


daattali:
I mentioned on the original issue where I submitted the reprex that this issue is fixed. And that's true. But this PR introduces a bigger bug: `lng` no longer gets exposed in the click dataframe.

Place this code inside the `observe({})`

```
cat(input$map_marker_click$lat, '\n')
cat(input$map_marker_click$lng)
```

Click on a marker, only the lat gets printed.

Maybe it's not this PR but rather some other commit between the CRAN version and right now.

schloerke:
Correct.  I had typed `lon` not `lng`.  Thank you @daattali !  Will fix in new PR.

daattali:
Ah it's a mistake I constantly make. There really should be a global
accepted convention for that shorthand!

On Jun 28, 2018 2:38 PM, "Barret Schloerke" <notifications@github.com>
wrote:

Correct. I had typed lon not lng. Thank you @daattali
<https://github.com/daattali> ! Will fix in new PR.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/rstudio/leaflet/pull/563#issuecomment-401132719>, or mute
the thread

schloerke:
@daattali Fixed in #566 . This is now merged in master.

PR #568, opened 2018-07-14
to make addMiniMap accessible within other packages

mtennekes:
Small fix to to make `addMiniMap` accessible within other packages.
See https://github.com/mtennekes/tmap/issues/231


schloerke:
@jcheng5 said no CLA needed

PR #592, opened 2018-10-30
fix incorrect library name in morefeatures.Rmd

richpauloo:
`htmlwidgets` (plural) is the library name, not `htmlwidget`

This is simply a typo fix in the instructional docs. No changes to codebase.


schloerke:
Thank you, @richpauloo !

PR #614, opened 2019-04-09
Added polyline offset plugin support with example data and script

fermm92:
## Pull Request

Hi,

I have implemented https://github.com/bbecquet/Leaflet.PolylineOffset which is great for visualizing directional road flows on road networks.  I was wondering if this would be the best way to share the work with the community. 

Essentially I modify the layers file and js sources to accept 'offset' as a variable and add new dependencies for the plug-in

Best regards,
Fernando


inst/htmlwidgets/leaflet.yaml
```diff
@@ -22,3 +22,7 @@ dependencies:
     version: 1.3.1 # match leaflet version
     src: "htmlwidgets/lib/rstudio_leaflet"
     stylesheet: rstudio_leaflet.css
+  - name: leaflet-polylineoffset
+    version: 1.1.1 # match leaflet version
+    src: "htmlwidgets/plugins/leaflet-polylineoffset"
+    stylesheet: leaflet.polylineoffset.js
```
trafficonese: It should be `script` instead of `stylesheet`. Otherwise the js-file would be added as css-link

scripts/offset_example.R
```diff
@@ -0,0 +1,6 @@
+# example
+corunaroads <- readRDS("./data/corunaroads.rds")
```
trafficonese: I am not able to install your PR because of the .rds file. Could you either save it as .rda file or maybe just create some lines from sampled coordinates.

> file 'corunaroads.rda' has magic number 'X'
>   Use of save versions prior to 2 is deprecated

Also I see that your example uses `readRDS("./data/corunaroads.rds")` but the PR has only ` data/corunaroads.rda`?

inst/htmlwidgets/leaflet.yaml
trafficonese: Isnt the **src** in `/lib` instead of `/plugin`?

R/plugin-polylineoffset.R
```diff
@@ -0,0 +1,10 @@
+leafletPolylineoffsetDependencies <- function() {
+  list(
+    htmltools::htmlDependency(
+      "leaflet-polylineoffset",
+      "1.1.1",
+      system.file("htmlwidgets/lib/leaflet-polylineoffset", package = "leafletfmm"),
```
trafficonese: It probably should be `package = "leaflet"`

R/layers.R
```diff
@@ -1128,6 +1128,7 @@ addPolylines <- function(
   fill = FALSE,
   fillColor = color,
   fillOpacity = 0.2,
+  offset = NULL,
```
trafficonese: I think new parameters should be added at the end to keep backward compatibility for positional arguments.

trafficonese: The documentation for `offset` is missing.

scripts/offset_example.R
fermumen: I found out that the `rda`  file was a badly renamed `rds` file hence the error. I've managed to fix it in the latest commit.

R/layers.R
fermumen: I've added the documentation for the parameter but I'm not sure what's the best way to update the `man` files. Is it just a case of using `devtools::document()` ?

fermumen:
@trafficonese I've gone through your comments hopefully you'll be able to install it and test it now.

R/layers.R
trafficonese: yes, or you can change the settings in RStudio under Build > More > Configure Build Tools. In the Panel under "Build Tools" you can click the "Configure" Button and check "Install and Restart".

trafficonese: If you want to leave the dataset "corunaroads" in the PR, then you need to document that aswell to pass the R checks. You can see the documentation for the other datasets in `/R/data.R`. Then document & reinstall and resubmit the changes.

Then you should pass the Travis Checks for R. I'm not sure about the Javascript Check.

fermumen:
I decided to do without `corunaroads` as it's probably a bit redundant. Somehow running `document()` has created a conflict on _map-shiny.Rd_. Any idea why?

trafficonese:
I think it is because of the new roxygen version. But you should be able to look at the conflicts, resolve them and merge the results.

fermumen:
@trafficonese  The R checks are ok but as you said we are not passing the Javascript one. I'm a bit out of my depth in JS, unfortunately. It is entirely possible that I haven't properly followed [these steps](https://github.com/rstudio/leaflet#development)? I'm not sure since I prepared the pull request in April last year and I don't remember 😅

trafficonese:
Unfortunately i'm no JS-Pro either and I dont know exactly what the checks are doing.
But did you use `yarn build` ? This should build a fresh `/htmlwidgest/leaflet.js`. Its not included in your PR so you probably skipped that part :)

schloerke:
Yes. Calling `yarn build` in the terminal will be necessary for the JS check to pass.

fermumen:
Success! Thanks @schloerke  and @trafficonese, the fresh yarn build did the trick.

trafficonese:
I have a little suggestion:

Could you include the `leafletPolylineoffsetDependencies` only if an offset is used?
Otherwise the dependency is always added, but not always needed/required.

So if offset is 0/NULL/NA, dont include the dependency.

fermumen:
For some reason, I'm getting some error when installing, but I think is entirely due to my machine:
`ERROR: dependency 'leaflet.providers' is not available for package 'leaflet'` 

I'll try to fix it but in the meanwhile, I haven't been able to install the package and test the new modifications.

fermumen:
It looks like the Polylineoffset dependency is being added to the maps regardless we use `leafletPolylineoffsetDependencies` or not. Any idea why?

AntonWijbenga:
Are issues with lines becoming circles or other erratic shapes solved in this version? The issues are discussed in the issue list of the bbecquet repository. For example here: [https://github.com/bbecquet/Leaflet.PolylineOffset/issues/1](https://github.com/bbecquet/Leaflet.PolylineOffset/issues/1)

I once adjusted a fork that had fixed such issues and made it compatible with the never version fo Leaflet (v2?). I use ShinyJS to include the .js source and then just use the offset parameter in the addPolylines function. Example: addPolylines(..., options=list(offset=3), ...) which works.

The JS file I use can be found here:
[https://traphx.maptm.nl/leaflet.polylineoffset_arw.js](https://traphx.maptm.nl/leaflet.polylineoffset_arw.js)

I have limited experience with package building and/or JavaScript. Otherwise I'd give it a go myself.

fermumen:
Hi @AntonWijbenga,

Those issues with erratic shapes exist in the version implemented here, particularly the circles problem would happen for very complex shapes. In the past I have managed to mitigate it by simplifying the shape with `rmapshaper` package. I haven't touched this PR in ages (two years) but if there is a version of the polyline offset that solves it you should be able to include it in the appropriate folder in `inst/htmlwodgets/lib/` to make it work.

ainefairbrother:
Hi, thanks for implementing this feature! I would really like to use it, but can't find any documentation regarding how to implement this. Could you please provide a gist? 

Thanks again, Aine

fermumen:
> Hi, thanks for implementing this feature! I would really like to use it, but can't find any documentation regarding how to implement this. Could you please provide a gist?
> 
> Thanks again, Aine

Hi! 

This was never merged to the main branch, and it was implemented a while ago. Not sure how easy it would be to make it work but try `remotes::install_github("rstudio/leaflet#614")` and perhaps use a CRAN mirror for Early 2020, you can look at the example in `scripts` from this PR. Alternatively if you can switch to python, the Folium team has this implemented [here](https://python-visualization.github.io/folium/latest/user_guide/plugins/polyline_offset.html)

Issue #786, opened 2022-03-21
OpenStreetMap tiles not rendering in some circumstances


jcheng5:
```r
library(leaflet)
leaflet() %>% addTiles()
```

The result works for me in RStudio or Safari, but not if I view it using Chrome (Version 99.0.4844.83, macOS). I get 403 errors for all of the map tiles.

PR #787, opened 2022-03-21
Fix OpenStreetMap tiles not rendering under Chrome

jcheng5:
Fix https://github.com/rstudio/leaflet/issues/786

Basically, the problem was from trying to use http URLs when possible. Just use https everywhere.

PR task list:
- [x] Update NEWS
- [ ] Add tests (where appropriate)

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -4,6 +4,8 @@ leaflet 2.1.0.9000
 BUG FIXES and IMPROVEMENTS
 * The default marker icon for `addMarkers` no longer worked, due to the CDN we were relying on apparently being retired. Fixed by pointing to a new CDN. (#782)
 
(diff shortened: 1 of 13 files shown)


schloerke (approved):
Url changes look great. 

I remember we did this because of Linux on IDE. 

If they are ok to have https, then all of this is awesome. If they can not visualize https in the IDE on Linux, then we will have problems.

jcheng5:
I think it was, like, Linux on Red Hat 5 😆

jcheng5:
I can at least confirm that it works on Ubuntu 20.04. If someone out there still needs http, I think it just means they will have to provide the URL manually.

PR #803, opened 2022-05-16
Added option to addProviderTiles() for user to specify .js file location

jaredlander:
When calling `addProviderTiles`, JavaScript is written to a temp file. This does not persist when using Quarto (but it has in the past with regular rmarkdown). So I added an option in `addProviderTile` (along with subsequently called functions) for the user to specify a location to write that JavaScript file. This fixed the problem for a presentation I am working on that has many called to `addProviderTiles`.

## Minimal reproducible example

```r
library(leaflet)

# this is the inst folder for leaflet
pop <- sf::read_sf('inst/csv/uspop2000.csv') |> 
  sf::st_as_sf(coords=c('Long', 'Lat'), remove=FALSE)

leaflet() |> 
  addProviderTiles('CartoDB.Voyager', dependencyLocation='~') |> 
  addCircles(data=pop)
```

PR task list:
- [X ] Update NEWS
- [ ] Add tests (where appropriate)
- [ X] Update documentation with `devtools::document()`

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: leaflet
 Type: Package
 Title: Create Interactive Web Maps with the JavaScript 'Leaflet' Library
-Version: 2.1.1
+Version: 2.1.2
 Authors@R: c(
       person("Joe", "Cheng", email = "joe@rstudio.com", role = c("aut", "cre")),
       person("Bhaskar", "Karambelkar", role = c("aut")),
--- a/NEWS
+++ b/NEWS
@@ -1,3 +1,9 @@
+leaflet 2.1.2
+--------------------------------------------------------------------------------
+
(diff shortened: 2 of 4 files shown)


jaredlander:
I'll admit this is not a pretty solution, but it got the job done. Maybe something nicer will be to automatically write somewhere other than `tempdir()`. Or perhaps include the JavaScript as an R object rather than a file, but I don't know if that's possible given the JavaScript that needs to be invoked.

jaredlander:
Turns out this problem also creates issues for regular rmarkdown documents that use caching. When using `addProviderTiles()` extensively last year this wasn't a problem, so there must have been a change somewhere in the past year that puts the downloaded javascript files somewhere (/tmp) that doesn't survive renders.

schloerke:
Closing in favor of #884

Issue #464, opened 2017-09-12
Please add scrollWheelZoom Leaflet option


behrman:
Have you ever been scrolling in a document and when you scrolled over a Leaflet window the map zoomed in or out? This can be very annoying.

Could you please implement the scrollWheelZoom Leaflet option.

I tried

    leaflet(options = leafletOptions(scrollWheelZoom = FALSE)) %>%

but this didn't work.

Many thanks,

Bill

bhaskarvk:
Take a look at `leaflet.extras:: suspendScroll()` function. It is designed for the exact kind of use case you mention. https://bhaskarvk.github.io/leaflet.extras/reference/sleep.html
For your case setting `hoverToWake=FALSE` will prevent the map from activating as you scroll and will require a click or a tap (on mobile devices) to be able to interact w/ the map.

behrman:
Bhaskar,

Many thanks for the terrific work you are doing with leaflet.extras!

Bill

VSzabo82:
Hi Bhaskar,

i discovered that the suspendScroll(hoverToWake = FALSE) option is not working anymore in the latest R and leaflet package version.

For example:
leaflet(width = "100%") %>%
  setView(0, 0, 1) %>%
  addTiles() %>%
  suspendScroll(hoverToWake = FALSE)

The map is still zooming in and out when i hover the mouse over it and start scrolling. Is there a possibility that you could fix this feature again? That would be perfect.

Thank you,
Viktoria

tcwilkinson:
suspendScroll is useful if version works (although seems not to be working in current version), but to permanently turn off scroll wheel (not just to prevent accident scrolls) is harder. 

Can I request the feature of an option `leaflet(options = leafletOptions(scrollWheelZoom = FALSE))` be included in base leaflet?

GibranMena:
> Hi Bhaskar,
> 
> i discovered that the suspendScroll(hoverToWake = FALSE) option is not working anymore in the latest R and leaflet package version.
> 
> For example:
> leaflet(width = "100%") %>%
> setView(0, 0, 1) %>%
> addTiles() %>%
> suspendScroll(hoverToWake = FALSE)
> 
> The map is still zooming in and out when i hover the mouse over it and start scrolling. Is there a possibility that you could fix this feature again? That would be perfect.
> 
> Thank you,
> Viktoria

I'm having the same issue.

nicofierrov:
> Hi Bhaskar,
> 
[...]

PR #827, opened 2022-10-31
fixed scrollWheelZoom to disable scroll zoom

byzheng:
Fix the bug to disable scroll zoom with option (scrollWheelZoom = FALSE) #464.


jcheng5:
Thank you @byzheng!

Issue #833, opened 2022-12-15
Please remove dependencies on **rgdal**, **rgeos**, and/or **maptools**


rsbivand:
This package depends on (depends, imports or suggests) **raster** and one or more of the retiring packages **rgdal**, **rgeos** or **maptools** (https://r-spatial.org/r/2022/04/12/evolution.html, https://r-spatial.org/r/2022/12/14/evolution2.html). Since **raster** `3.6.3`, all use of external FOSS library functionality has been transferred to **terra**, making the retiring packages very likely redundant. It would help greatly if you could remove dependencies on the retiring packages as soon as possible.

My guess is that dropping **rgdal** and **rgeos** and replacing them by **terra**, conditioning **raster** on `>= 3.6.3` will resolve this (untried). You use `rgeos::createPolygonsComment`, but should be able to work around that with either **sf** or **terra**.

jcheng5:
Thanks @rsbivand. With regard to `rgeos::createPolygonsComment`, it looks to me like we use this for **sp** `Polygons` objects whenever they're missing comment data indicating, for each hole, what non-hole they belong to.

> `createPolygonsComment` attempts to create a valid comment for a `Polygons` object by assessing which polygons contain a given hole (using [gContains](http://127.0.0.1:14629/help/library/rgeos/help/gContains)). Ownership is assigned to the smallest polygon (by area) that contains the given hole.

Is there some higher-level functionality in **sf** or **terra** that we can use for this, or were you imagining a reimplementation using just `st_contains` and `st_area`?

rsbivand:
I'll create a reprex in https://github.com/r-spatial/evolution https://github.com/r-spatial/evolution/issues/9 and explore. It should no longer be a big problem, if for example ESRI Shapefiles are read through the GDAL driver where GDAL was built linking to GEOS, the driver itself restores Simple Features specs even via `rgdal::readOGR()`. It was a problem before about 2010, and still would be for serialised `Polygons` objects pre-dating those changes, and for other manually created objects.

jcheng5:
2010! Wow...

What if we detect Polygons with 1) missing comment, 2) holes, 3) more than one island, and throw a warning (or even error)? Technically breaking backward compatibility I suppose, but this seems like such an extreme edge case.

rsbivand:
[...]

PR #837, opened 2022-12-22
Remove outdated dependencies; use testthat-3e

jcheng5:
See #833

The code change in normalize-sp.R is assuming that https://github.com/r-spatial/sf/pull/2069 gets merged. If not, either we'll need similar code, or else just warn/error that we don't handle this edge case. In either case, I need to write a test that exercises that code path.

PR task list:
- [x] Update NEWS
- [x] Add tests (where appropriate)


jcheng5:
@gadenbuie I added the unit test we discussed today; this is now ready to go. However, the rare case we discussed (sp objects that have holes but are missing the comment that indicates hole ownership) only works with sf 1.0-10, which is not yet on CRAN. If this case is encountered while the CRAN version of sf is installed, leaflet will report an error telling you to upgrade sf to 1.0-10 or above. The tests will pass either way, as I don't attempt the unit test if the wrong version of sf is installed.

I don't think we have to wait for sf 1.0-10 to hit CRAN to merge this PR, but I do think we should wait for that before we submit leaflet to CRAN.

PR #924, opened 2024-07-25
feat: Expand available `{viridis}` palettes in `colorNumeric()`

jack-davison:
Since #897 `{leaflet}` uses `{viridisLite}`.

`colorNumeric()` only allows users to easily use four of the eight palettes available through `{viridisLite}`.

This PR allows the other four - turbo, mako, cividis, and rocket.

Slightly selfish PR in that I do like "turbo" for a not totally dreadful rainbow palette! Also I can't see a reason why only four of the eight palettes should be allowed / documented.

## Minimal reproducible example

``` r
devtools::load_all()
#> ℹ Loading leaflet

pal <- colorNumeric("turbo", breweries91$founded)

leaflet(breweries91) |>
  addTiles() |>
  addCircleMarkers(color = ~ pal(founded)) |>
  addLegend(pal = pal,
            values = breweries91$founded)
```

<sup>Created on 2024-07-25 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

PR task list:
- [x] Update NEWS

No current tests for different viridisLite palettes, so no tests added.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -81,5 +81,5 @@ Config/testthat/edition: 3
 Config/Needs/website: dplyr, geojsonio, ncdf4, tidyverse/tidytemplate
 Encoding: UTF-8
 LazyData: true
-RoxygenNote: 7.3.1
+RoxygenNote: 7.3.2
 Roxygen: list(markdown = TRUE)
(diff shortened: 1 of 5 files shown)


schloerke:
Thank you!

